PR #3, opened 2019-05-03
Attach an idempotency key to each event when its generated

dewski:
By attaching an idempotency key to each event when it is generated we can ensure that if the same event is published more than once for any reason WorkOS can safely ignore writing the new entry to the Audit Log.


marktran:
👍 Corresponding changes on the API-side are deployed

marktran:
@dewski Just fixed some merge conflicts. Does this look good to merge?

PR #4, opened 2019-05-03
Configurable API Key

dewski:
This PR allows clients to configure the API key by using the `WORKOS_API_KEY` environment variable or through other means like secret managers.

Ideally this can be done using a configuration struct in the future when there is potentially more to configure, but for now keep it simple. It shouldn't require an API change should there be a configuration struct as the client will provide a default client it will set the API token to.


marktran:
@sdalezman Would this work for your use case?

sdalezman:
looks great!

Issue #6, opened 2019-05-17
Add auto pagination


marktran:
Similar to https://stripe.com/docs/api/pagination/auto

PR #7, opened 2019-05-19
Implementing auto pagination

maxchehab:
#6 

Auto pagination easily handles fetching large lists of events without having to manually paginate results and perform subsequent requests. Optional parameters `Limit`, `StartingAfter`, and `EndingBefore` can be provided to query. If both `StartingAfter` and `EndingBefore` are provided, the client will default to using only `EndingBefore`.

```go
i := events.List(auditlog.ListRequestParams{})
for i.Next() {
    event := i.Event()
    log.Println(event.ID)
}
```


auditlog/api.go
```diff
@@ -16,174 +14,31 @@ const (
...
+// Get exucutes a get request to a provided resource
+func Get(path string) (*http.Response, error) {
 	client := http.Client{
 		Timeout: 10 * time.Second,
 	}
 
 	endpoint := os.Getenv("WORKOS_ENDPOINT")
 	if endpoint == "" {
-		endpoint = "https://api.workos.com"
+		endpoint = "https://max.workos.dev"
```
marktran: ☝️

auditlog/api.go
```diff
@@ -16,174 +14,31 @@ const (
...
 	}
 
 	route := fmt.Sprintf("%s%s", endpoint, path)
+	log.Println(route)
```
marktran: Was this just for debugging?

marktran:
@dewski would love to get your feedback on this

auditlog/api.go
```diff
@@ -205,3 +60,47 @@ func get(path string) (*http.Response, error) {
...
+	if p.Limit <= 0 {
+		return 10
+	}
+	return p.Limit
+}
+
+type ListMeta struct {
+	HasMore    bool   `json:"has_more"`
+	TotalCount uint32 `json:"total_count"`
+	URL        string `json:"url"`
```
dewski: Could be helpful to document what these fields may contain. What's URL in this example?

auditlog/events/client.go
```diff
@@ -0,0 +1,92 @@
...
+
+const (
+	userAgent  = "workos-go/0.1"
+	eventsPath = "/events"
+)
+
+// List returns a list of events.
+func List(params auditlog.ListRequestParams) *Iter {
+	return &Iter{auditlog.GetIter(params, func(params auditlog.ListRequestParams) ([]interface{}, auditlog.ListMeta, error) {
+		list := auditlog.EventList{}
```
dewski: `list` is overwritten on the next line:

```suggestion
```

auditlog/events/client.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+const (
+	userAgent  = "workos-go/0.1"
+	eventsPath = "/events"
+)
+
+// List returns a list of events.
+func List(params auditlog.ListRequestParams) *Iter {
+	return &Iter{auditlog.GetIter(params, func(params auditlog.ListRequestParams) ([]interface{}, auditlog.ListMeta, error) {
```
dewski:
```suggestion
	return &Iter{auditlog.GetIter(params, func(params auditlog.ListRequestParams) ([]*auditlog.Event, auditlog.ListMeta, error) {
```

auditlog/events/client.go
```diff
@@ -0,0 +1,92 @@
...
+		list := auditlog.EventList{}
+		list, err := FindAll(params)
+
+		ret := make([]interface{}, len(list.Data))
+		for i, v := range list.Data {
+			ret[i] = v
+		}
+
+		return ret, list.ListMeta, err
```
dewski: It wasn't immediately clear why `[]interface{}` is being returned instead of `*auditlog.Event`, but based on the implementation it looks like you're building an array of `*auditlog.Event` structs.

auditlog/api.go
maxchehab: `URL` is the endpoint which the data belongs to. This is useful for nested lists where one may want to paginate the child's list. For example..
```json
GET /foo
{
    "object": "foo",
    "url": "/foo",
    "has_more": true,
    "data": [{
        "object": "foo",
        "id": "foo_01DAYKK3T3YCX7F2FM9J4ZCRB1",
        "bars": {
            "object": "bar",
            "url": "/foo/foo_01DAYKK3T3YCX7F2FM9J4ZCRB1/bar",
            "has_more": false,
            "data": [{
                "object": "bar",
                "id": "bar_01DAYKK3T3YCX7F2FM9J4ZCRB1",
                "value": "foobar"
            }]
        }
    }]
}
```

main.go
```diff
@@ -64,9 +64,8 @@ func main() {
 	})
 
 	http.HandleFunc("/events", func(w http.ResponseWriter, req *http.Request) {
-		resp, err := auditlog.FindAll(auditlog.EventsRequestParams{
-			End:    time.Now(),
-			Action: "user.login",
+		resp, err := events.FindAll(auditlog.ListRequestParams{
```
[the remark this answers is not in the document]
maxchehab: `Iter` and `ListRequestParams` are both reusable structs that can be used to paginate over varying data. By not including `ListRequestParams` to the `events` package we can reuse it if there is another entity that requires pagination.

auditlog/events/client.go
maxchehab: As I mentioned before, I tried to make `Iter` reusable with other entities that require pagination. If there is a new entity, all the developer has to do is implement a new `Query` function.

maxchehab:
@dewski Thank you for your feedback, it's very appreciated. 

Just for context, I unfortunately saw the this sdk's implementation of pagination after I implemented it on the api. If you would like to discuss paginating by time vs index I am happy to have a conversation.

> My one concern with these changes would be that it may not be optimizing for the most important needs an integrator may have with WorkOS. This auto paginated API implementation removes the ability to search by action or time for example.

That being said, the ability to search by action or time is not supported by WorkOS api. This is a planned feature, but I thought that since it was not available then it should not be included in the sdk as of now. This is the next story I am going to work on.

auditlog/api.go
```diff
@@ -16,174 +14,31 @@ const (
 )
 
 var (
-	apiKey = ""
+	ApiKey = ""
```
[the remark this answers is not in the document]
maxchehab: I think at some point I was accessing `apiKey` from the `event` package and didn't realize it was no longer required.

maxchehab:
@dewski I am not sure if you have access, but it may be worth your time to check this PR that implements pagination. https://github.com/workos-inc/workos/pull/46

auditlog/event.go
```diff
@@ -50,9 +50,16 @@ const (
...
+// EventList represents a set of Audit Log events returned from WorkOS.
+type EventList struct {
+	ListMeta
+	Data []*Event `json:"data"`
+}
+
 // Event represents the structure of a Audit Log event with all the necessary
 // metadata needed to describe an event properly.
 type Event struct {
+	ID         string     `json:"id"`
```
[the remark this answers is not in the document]
maxchehab: `ID` is now required to paginate entities. We also have implemented the route `/events/:id` where the client is required to know the `ID` of an event.

marktran:
Thanks @dewski!

Filtering is on-deck this week and we definitely want pagination to work in conjunction with that.

marktran:
@dewski What do you think about merging this?

FYI, we ended up designing pagination and filtering a bit different from the original spec (@maxchehab is just updated it in Notion).

maxchehab:
Now uses `before` and `after` vocabulary.

main.go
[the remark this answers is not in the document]
dewski: It feels a little soon for an abstraction, might I suggest having the `events` package expose `ListRequestParams`? Once there are more resources to iterate through it would make more sense to DRY it up.

auditlog/iter.go
```diff
@@ -0,0 +1,67 @@
+package auditlog
+
+import (
+	"reflect"
+)
+
+// Query is the function used to get a page listing.
+type Query func(ListRequestParams) ([]interface{}, ListMeta, error)
+
+// EventItr represents and iterative event
```
dewski:
```suggestion
// Itr provides a convenient interface for reading a list of interfaces. It stops when there are no more interfaces to iterate over.
```

auditlog/events/client.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// FindAll returns a paginated set of Audit Log entries matching the search
+// query.
+func FindAll(params auditlog.ListRequestParams) (auditlog.EventList, error) {
+	path := eventsPath
+	q := url.Values{}
+
+	q.Add("limit", strconv.Itoa(params.GetLimit()))
+	q.Add("after", params.GetStartingAfter())
```
dewski: Since a string is always provided even if there is no `after` we should probably check it isn't empty.

```suggestion
	if params.GetStartingAfter() != "" {
		q.Add("after", params.GetStartingAfter())
	}
```

auditlog/events/client.go
```diff
@@ -0,0 +1,89 @@
...
+	q.Add("before", params.GetEndingBefore())
```
dewski: Similar to the `after` feedback here.

auditlog/events/client.go
```diff
@@ -0,0 +1,89 @@
...
+	query := q.Encode()
+	if query != "" {
+		path = fmt.Sprintf("%s?%s", path, query)
+	}
+
+	resp, err := auditlog.Get(path)
+	if err != nil {
+		return auditlog.EventList{}, err
+	}
+
```
dewski: To match the function below.

```suggestion
```

auditlog/events/client.go
```diff
@@ -0,0 +1,89 @@
...
+			ret[i] = v
+		}
+
+		return ret, list.ListMeta, err
+	})}
+}
+
+// FindAll returns a paginated set of Audit Log entries matching the search
+// query.
+func FindAll(params auditlog.ListRequestParams) (auditlog.EventList, error) {
```
dewski: Feels like the event package should know about `EventList`.

auditlog/api.go
```diff
@@ -205,3 +58,52 @@ func get(path string) (*http.Response, error) {
...
+
+// ListRequestParams allows you to confire FindAll or List request to paginate
+// any entries after & before a given index.
+type ListRequestParams struct {
+	After  string
+	Before string
+	Limit  int
+}
+
+func (p ListRequestParams) GetStartingAfter() string {
```
dewski:
```suggestion
func (p ListRequestParams) GetAfter() string {
```

auditlog/api.go
```diff
@@ -205,3 +58,52 @@ func get(path string) (*http.Response, error) {
...
+	if p.Before != "" && p.After != "" {
+		return ""
+	}
+	return p.After
+}
+
+func (p ListRequestParams) GetEndingBefore() string {
```
dewski:
```suggestion
func (p ListRequestParams) GetBefore() string {
```

auditlog/events/client.go
dewski: Since `interface{}` is the return type you should be able to just pass `list.Data` directly since it matches.

auditlog/event.go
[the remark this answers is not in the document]
dewski: There is an `ID` field on `EventResponse` making it redundant. `EventList` is currently depending on `Event` but that doesn't include the location fields or other fields that happen after the event is processed. Should `EventList` instead have `[]*EventResponse`?

auditlog/iter.go
```diff
@@ -0,0 +1,67 @@
...
+type Iter struct {
+	values  []interface{}
+	current interface{}
+	err     error
+	params  ListRequestParams
+	meta    ListMeta
+	query   Query
+}
+
+// Next returns the next event in the set
```
dewski: If the Iter interface is for any resource maybe it should be:
```suggestion
// Next returns the next value in the set
```

auditlog/iter.go
```diff
@@ -0,0 +1,67 @@
...
+		params: params,
+		query:  query,
+	}
+
+	iter.getPage()
+
+	return iter
+}
+
+func listItemID(x interface{}) string {
```
dewski: I think it'd make more sense to take advantage of go's interfaces so each each list item would have to implement the following function:

```go
type Iterable interface {
  GetID() string
}
```

auditlog/events/client.go
dewski: Is it not possible to return `list.Data` here? Since the elements in `list.Data` all implement the Iterable interface it should work as:

```suggestion
		return list.Data, list.ListMeta, err
```

auditlog/event.go
```diff
@@ -129,6 +130,11 @@ func NewEventWithMetadata(action Action, actionType ActionType, metadata map[str
 	return event, nil
 }
 
+// GetID gets the Event's ID
```
dewski: I don't think this `Event` struct should know anything about the `ID` since the client cannot set it before sending it to WorkOS. It seems like `EventResponse` is the only struct that can know about the `ID`.

auditlog/iter.go
```diff
@@ -0,0 +1,63 @@
+package auditlog
+
+type Iterable interface {
+	GetID() string
+}
+
+// Query is the function used to get a page listing.
+type Query func(ListRequestParams) ([]Iterable, ListMeta, error)
```
dewski: I didn't catch this in previous reviews, but I'm wondering if `Query` is a little generic. What do you think about:
```suggestion
type IterableQuery func(ListRequestParams) ([]Iterable, ListMeta, error)
```

auditlog/iter.go
```diff
@@ -0,0 +1,63 @@
...
+	it.values = it.values[1:]
+
+	return true
+}
+
+func (it Iter) Current() Iterable {
+	return it.current
+}
+
+func (it *Iter) getPage() {
```
dewski: For style, lets move the exported functions above the unexported functions.

maxchehab:
@dewski 
> Is it not possible to return list.Data here? Since the elements in list.Data all implement the Iterable interface it should work as:

Sorry, I should have replied directly to this earlier.
Essentially, no. Go doesn't let you convert a slice of types to a slice of interfaces implicitly. [Stephen Weinberg's comment explains this in detail](https://stackoverflow.com/questions/12753805/type-converting-slices-of-interfaces-in-go/12754757#12754757).

auditlog/event.go
maxchehab: Correct, I think I will move `EventResponse` to `event.go` and change its declaration to the following. 

```go
type EventResponse struct {
	Event
	ID              string `json:"id"`
	Object          string `json:"object"`
	LocationCity    string `json:"location_city"`
	LocationState   string `json:"location_state"`
	LocationCountry string `json:"location_country"`
	AppID           string `json:"app_id"`
}
```
`EventResponse` also now implements the `Iterable` interface.

README.md
```diff
@@ -323,3 +323,15 @@ http.HandleFunc("/login", func(w http.ResponseWriter, req *http.Request) {
   }
 }
 ```
+
+## Paginating Events
+
+This client provides auto pagination when querying events. This feature easily handles fetching large lists of events without having to manually paginate results and perform subsequent requests. Optional parameters `Limit`, `StartingAfter`, and `EndingBefore` can be provided to query. If both `StartingAfter` and `EndingBefore` are provided, the client will default to using only `EndingBefore`.
```
marktran:
```suggestion
This client provides auto pagination when querying events. This feature easily handles fetching large lists of events without having to manually paginate results and perform subsequent requests. Optional parameters `Limit`, `After`, and `Before` can be provided to query. If both `After` and `Before` are provided, the client will default to using only `Before`.
```

README.md
```diff
@@ -323,3 +323,15 @@ http.HandleFunc("/login", func(w http.ResponseWriter, req *http.Request) {
...
+
+```go
+  i := events.List(auditlog.ListRequestParams{Limit: 20, EndingBefore: "evt_01DARZVVM933M93J6XREKWS436"})
```
marktran:
```suggestion
  i := events.List(auditlog.ListRequestParams{Limit: 20, Before: "evt_01DARZVVM933M93J6XREKWS436"})
```

auditlog/api.go
```diff
@@ -205,3 +58,39 @@ func get(path string) (*http.Response, error) {
 
 	return resp, nil
 }
+
+// ListRequestParams allows you to confire FindAll or List request to paginate
```
marktran:
```suggestion
// ListRequestParams allows you to configure FindAll or List request to paginate
```

PR #8, opened 2019-05-20
Added additional channel to access posted event

maxchehab:
This lets the integrator immediately see the event after it is published to the Audit Log.
For example:

```go
u := user{
	Email:      "me@domain.com",
	DatabaseID: 1,
}

event := auditlog.NewEventWithHTTP("user.login", auditlog.Create, req)
event.SetActor(u)
event.SetGroup(u)
event.SetTarget(u)

errCh, eventCh := event.Publish()
err := <-errCh
newEvent := <-eventCh
if err != nil {
	// Call out to sentry
	fmt.Fprintf(w, err.Error())
	return
}

body, _ := json.Marshal(newEvent)
fmt.Fprintf(w, string(body))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ The resulting event being sent to WorkOS looks like:
 
 The time the event occured is automatically populated for you when the event is created.
 
-All events are published to WorkOS asyncronously by default. `auditlog.Publish` returns an error channel for you so you can wait for a response from WorkOS should you need a blocking operation.
+All events are published to WorkOS asyncronously by default. `auditlog.Publish` returns an event and error channel for you so you can wait for a response from WorkOS should you need a blocking operation.
 
 ```go
 user := User{
@@ -88,11 +88,13 @@ event.SetGroup(organization)
 event.SetActor(user)
 event.SetTarget(user)
 event.SetLocation("1.1.1.1")
-ch := event.Publish()
-err := <-ch
+errCh, eventCh := event.Publish()
+err := <-errCh
+postedEvent := <-eventCh
 if err != nil {
   fmt.Printf("Had a problem writing the event: %q %q\n", event, err)
 }
+fmt.Println(postedEvent.ID)
 ```
 
 ## Configuring An Auditable Interface
@@ -150,7 +152,9 @@ event.AddMetadata(map[string]interface{}{
 	"body_was": bodyWas,
 	"body": comment.Body,
 })
-err := event.Publish()
+errCh, eventCh := event.Publish()
+err := <-errCh
+postedEvent := <-eventCh
 if err != nil {
 	fmt.Printf("Had a problem writing the event: %q %q\n", event, err)
 }
@@ -205,7 +209,9 @@ event.SetLocation("1.1.1.1")
 event.AddMetadata(map[string]interface{}{
 	"parent_tweet": tweet.ParentTweet,
 })
-err := event.Publish()
+errCh, eventCh := event.Publish()
+err := <-errCh
+postedEvent := <-eventCh
 if err != nil {
 	fmt.Printf("Had a problem writing the event: %q %q\n", event, err)
 }
@@ -294,7 +300,9 @@ http.HandleFunc("/login", func(w http.ResponseWriter, req *http.Request) {
 	event.SetGroup(user)
 	event.SetActor(user)
 	event.SetTarget(user)
-	err := event.Publish()
+	errCh, eventCh := event.Publish()
+  err := <-errCh
+  postedEvent := <-eventCh
 	if err != nil {
 		fmt.Printf("Had a problem writing the event: %q %q\n", event, err)
 	}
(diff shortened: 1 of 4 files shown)


auditlog/event.go
```diff
@@ -199,31 +200,34 @@ func (e Event) addMetadata(key string, value interface{}) error {
 }
 
 // Publish delivers the event to WorkOS asyncronously.
-func (e Event) Publish() chan error {
+func (e Event) Publish() (chan error, chan *Event) {
```
dewski: Errors are expected to be the second return value:

```suggestion
func (e Event) Publish() (chan *Event, chan error) {
```

README.md
```diff
@@ -294,7 +300,9 @@ http.HandleFunc("/login", func(w http.ResponseWriter, req *http.Request) {
 	event.SetGroup(user)
 	event.SetActor(user)
 	event.SetTarget(user)
-	err := event.Publish()
+	errCh, eventCh := event.Publish()
+  err := <-errCh
```
dewski:
```suggestion
	err := <-errCh
```

README.md
```diff
@@ -294,7 +300,9 @@ http.HandleFunc("/login", func(w http.ResponseWriter, req *http.Request) {
...
+  postedEvent := <-eventCh
```
dewski:
```suggestion
	postedEvent := <-eventCh
```

dewski:
Have some questions more from an API standpoint that I likely don't have full context on:

- Will WorkOS API always provide a full event after it's accepted through the API endpoint? For example, will it ever go into a queue, and if so, will the ID be assigned before that happens?
- The current event channel exposed won't return the events location information. It may be slightly confusing to have multiple shapes of an event at different stages.

maxchehab:
>Will WorkOS API always provide a full event after it's accepted through the API endpoint? For example, will it ever go into a queue, and if so, will the ID be assigned before that happens?
I'm not sure; @marktran do you have any knowledge on this?

>The current event channel exposed won't return the events location information. It may be slightly confusing to have multiple shapes of an event at different stages.
We should add the location parameter. I don't think we should hide any part of the event entity besides internal properties like `deleted_at` and such.

dewski:
> We should add the location parameter. I don't think we should hide any part of the event entity besides internal properties like `deleted_at` and such.

Unless the integrator can set the location information themselves I'd think there would be two structs, one for the user generated event and another for the server supplied event. I'd defer to @marktran on the explicit contracts between the server and client.

maxchehab:
Do you mean,if the the integrator can update the location information after the event is posted or before? AFAIK the WorkOS allows the [location parameter to be set per post request](https://docs.workos.com/reference#events).

maxchehab:
I agree that we should not let the integrator set an event's ID. We could make this property private and read-only through a getter.

dewski:
> Do you mean,if the the integrator can update the location information after the event is posted or before? AFAIK the WorkOS allows the [location parameter to be set per post request](https://docs.workos.com/reference#events).

I can't access that page. The integrator can set the `location` before the event is sent and then WorkOS processes the IPv4 or IPv6 address to populate the event with the country, state, city, etc. If that isn't the case anymore you can ignore this feedback 😄

dewski:
> I agree that we should not let the integrator set an event's ID. We could make this property private and read-only through a getter.

Sorry if I wasn't super clear with what I meant, I can try to explain another way.

The `Event` struct is the struct the integrator builds and delivers to WorkOS. I didn't originally add the `ID`, `LocationCountry`, `LocationState`, `LocationCity`, `Object`, and `AppID` fields to the `Event` struct to create an explicit struct that only contained the fields available to the integrator to configure.

When requesting events it'd return an `EventResponse` as at that point `AppID`, `ID`, `LocationCountry`, `LocationState`, and `LocationCity` would be generated. By adding `ID` to `Event` here it implies that WorkOS will write the event to a database right away so the `ID` is known, but will WorkOS ever write these to a queue to process the event? Will WorkOS generate the final ID before it's queued? Will it not be queued? The idempotency key is sent so the same event isn't stored twice. Will geolocation happen in the request lifecycle when the event is originally submitted? Just some questions that could help with the implementation here.

I'm also curious on what returning the event in the `Publish` call enables the integrator to do that they can't already do? I ask because all the information is already available since they generated the event with the action, CRUD, targets, metadata, timestamp, etc.

marktran:
> - Will WorkOS API always provide a full event after it's accepted through the API endpoint? For example, will it ever go into a queue, and if so, will the ID be assigned before that happens?
> - The current event channel exposed won't return the events location information. It may be slightly confusing to have multiple shapes of an event at different stages.

Good points, @dewski. We've discussed some of this internally and I'll share a doc that outlines some of it.
- We're always going to return an ID, even with a processing pipeline for enriching data, indexing, etc.
- Agreed, having multiple shapes for enriched/non-enriched events would be confusing. I think we might end up having a `expand` parameter that you pass in to expand the event with enriched data (somewhat similar to Stripe: https://stripe.com/docs/api/expanding_objects). Need to think about this some more and document it.

maxchehab:
@dewski @marktran I'm closing this PR due to the decision to not include the event data along with the `POST` response.

marktran:
👍 Yup, ID is removed from the response.

PR #12, opened 2019-10-22
Auditlog worker

maxence-charriere:
## Summary
This PR provides a Go implementation of our Audit Log SDK based on the one we already have.
This implementation provides:
- Possibility to use a custom `http.Client`
- Possibility to use a custom JSON encoder

```go
package main

import "github.com/workos-inc/workos-go/pkg/auditlog"

func main() {
    auditlog.SetAPIKey("my_api_key")

    // Wherever you need to publish an audit log event:
    err := auditlog.Publish(ctx.Background(), auditlog.Event{
        Action:     "document.viewed",
        ActionType: auditlog.Create,
        ActorName:  "Jairo Kunde",
        ActorID:    "user_01DGZ0FAXN978HCET66Q98QMTQ",
        Group:      "abstract.com",
        Location:   "55.27.223.26",
        OccurredAt: time.Now(),
        TargetName: "central.class",
        TargetID:   "doc_01DGZ0FAXP4HA4X0BVFKS0ZH4Y",
    })
    if err != nil {
        // Handle error.
    }
}
```

## Tests
- unit tested
- tested on a side project

Screen Shot 2019-10-22 at 9 22 06 PM

### DIFF
--- a/pkg/auditlog/README.md
+++ b/pkg/auditlog/README.md
@@ -1,0 +1,38 @@
+# WorkOS
+
+A WorkOS client for Go applications in your organization to control and monitor
+the access of information within your organization.
+
+## Install
+
+You can install the WorkOS Go client in your local environment by running:
+
+```sh
+go get -u github.com/workos-inc/workos-go/pkg/auditlog
+```
+
+## How it works
+
+```go
+package main
+
+import "github.com/workos-inc/workos-go/auditlog"
+
+func main() {
+    alog.SetAPIKey("my_api_key")
+    defer alog.Close()
+
+    // Wherever you need to publish an audit log event:
+    alog.Publish(alog.Event{
+        Action:     "document.viewed",
+        ActionType: "r",
+        ActorName:  "Jairo Kunde",
+        ActorID:    "user_01DGZ0FAXN978HCET66Q98QMTQ",
+        Group:      "abstract.com",
+        Location:   "55.27.223.26",
+        OccurredAt: time.Now(),
+        TargetName: "central.class",
+        TargetID:   "doc_01DGZ0FAXP4HA4X0BVFKS0ZH4Y",
+    })
+}
+```
--- a/pkg/auditlog/README.md
+++ b/pkg/auditlog/README.md
@@ -24,7 +24,7 @@ func main() {
     // Wherever you need to publish an audit log event:
     auditlog.Publish(auditlog.Event{
         Action:     "document.viewed",
-        ActionType: "r",
+        ActionType: auditlog.Create,
         ActorName:  "Jairo Kunde",
         ActorID:    "user_01DGZ0FAXN978HCET66Q98QMTQ",
         Group:      "abstract.com",
--- a/pkg/auditlog/README.md
+++ b/pkg/auditlog/README.md
@@ -19,10 +19,9 @@ import "github.com/workos-inc/workos-go/pkg/auditlog"
 
 func main() {
     auditlog.SetAPIKey("my_api_key")
-    defer auditlog.Close()
 
     // Wherever you need to publish an audit log event:
-    auditlog.Publish(auditlog.Event{
+    err := auditlog.Publish(ctx.Background(), auditlog.Event{
         Action:     "document.viewed",
         ActionType: auditlog.Create,
         ActorName:  "Jairo Kunde",
@@ -33,5 +32,8 @@ func main() {
         TargetName: "central.class",
         TargetID:   "doc_01DGZ0FAXP4HA4X0BVFKS0ZH4Y",
     })
+    if err != nil {
+        // Handle error.
+    }
 }
 ```
--- a/pkg/auditlog/README.md
+++ b/pkg/auditlog/README.md
@@ -13,7 +13,7 @@ go get -u github.com/workos-inc/workos-go/pkg/auditlog
 ```go
 package main
 
-import "github.com/workos-inc/workos-go/auditlog"
+import "github.com/workos-inc/workos-go/pkg/auditlog"
 
 func main() {
     alog.SetAPIKey("my_api_key")
--- a/pkg/auditlog/README.md
+++ b/pkg/auditlog/README.md
@@ -1,4 +1,4 @@
-# alog
+# auditlog
 
 A Go package to asynchronously send audit logs events to WorkOS.
 
@@ -16,11 +16,11 @@ package main
 import "github.com/workos-inc/workos-go/pkg/auditlog"
 
 func main() {
-    alog.SetAPIKey("my_api_key")
-    defer alog.Close()
+    auditlog.SetAPIKey("my_api_key")
+    defer auditlog.Close()
 
     // Wherever you need to publish an audit log event:
-    alog.Publish(alog.Event{
+    auditlog.Publish(auditlog.Event{
         Action:     "document.viewed",
         ActionType: "r",
         ActorName:  "Jairo Kunde",
--- a/pkg/auditlog/README.md
+++ b/pkg/auditlog/README.md
@@ -1,11 +1,8 @@
-# WorkOS
+# alog
 
-A WorkOS client for Go applications in your organization to control and monitor
-the access of information within your organization.
+A Go package to asynchronously send audit logs events to WorkOS.
 
 ## Install
-
-You can install the WorkOS Go client in your local environment by running:
 
 ```sh
 go get -u github.com/workos-inc/workos-go/pkg/auditlog
--- a/pkg/auditlog/README.md
+++ b/pkg/auditlog/README.md
@@ -1,5 +1,7 @@
 # auditlog
 
+[![godoc](https://godoc.org/github.com/workos-inc/workos-go/pkg/auditlog?status.svg)](https://godoc.org/github.com/workos-inc/workos-go/pkg/auditlog)
+
 A Go package to asynchronously send audit logs events to WorkOS.
 
 ## Install
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -82,7 +82,7 @@ func defaultLocation(location string) string {
 
 func defaultTime(t time.Time) time.Time {
 	if t == (time.Time{}) {
-		t = time.Now()
+		t = time.Now().UTC()
 	}
 	return t
 }
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -1,4 +1,23 @@
 // Package alog is a package to send audit logs events to WorkOS.
+//
+// Example:
+//   func main() {
+//       alog.SetAPIKey("my_api_key")
+//       defer alog.Close()
+//
+//       // Wherever you need to publish an audit log event:
+//       alog.Publish(alog.Event{
+//           Action:     "document.viewed",
+//           ActionType: "r",
+//           ActorName:  "Jairo Kunde",
+//           ActorID:    "user_01DGZ0FAXN978HCET66Q98QMTQ",
+//           Group:      "abstract.com",
+//           Location:   "55.27.223.26",
+//           OccurredAt: time.Now(),
+//           TargetName: "central.class",
+//           TargetID:   "doc_01DGZ0FAXP4HA4X0BVFKS0ZH4Y",
+//       })
+//   }
 package alog
 
 import (
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -3,10 +3,9 @@
 // Example:
 //   func main() {
 //       auditlog.SetAPIKey("my_api_key")
-//       defer auditlog.Close()
 //
 //       // Wherever you need to publish an audit log event:
-//       auditlog.Publish(auditlog.Event{
+//       err := auditlog.Publish(context.Background(), auditlog.Event{
 //           Action:     "document.viewed",
 //           ActionType: auditlog.Create,
 //           ActorName:  "Jairo Kunde",
@@ -17,6 +16,9 @@
 //           TargetName: "central.class",
 //           TargetID:   "doc_01DGZ0FAXP4HA4X0BVFKS0ZH4Y",
 //       })
+//       if err != nil {
+//           // Handle error.
+//       }
 //   }
 package auditlog
 
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -33,7 +33,29 @@ var (
 		Endpoint:  "https://api.workos.com/events",
 		QueueSize: 512,
 	}
+
+	currentLocation string
 )
+
+func init() {
+	currentLocation, _ = os.Hostname()
+}
+
+// SetAPIKey sets the WorkOS API key to use when using Publish.
+func SetAPIKey(k string) {
+	DefaultPublisher.APIKey = k
+}
+
+// Publish publishes the given events.
+func Publish(events ...Event) {
+	DefaultPublisher.Publish(events...)
+}
+
+// Close stops publishings audit log events and releases allocated resources.
+// It waits for pending events to be sent before returning.
+func Close() {
+	DefaultPublisher.Close()
+}
 
 // Event represents an Audit Log event.
 type Event struct {
@@ -51,18 +73,9 @@ type Event struct {
 	indempotencyKey string
 }
 
-// SetAPIKey sets the WorkOS API key to use when using Publish.
-func SetAPIKey(k string) {
-	DefaultPublisher.APIKey = k
-}
-
-// Publish publishes the given events.
-func Publish(events ...Event) {
-	DefaultPublisher.Publish(events...)
-}
-
-// Close stops publishings audit log events and releases allocated resources.
-// It waits for pending events to be sent before returning.
-func Close() {
-	DefaultPublisher.Close()
+func defaultLocation(location string) string {
+	if location == "" {
+		location = currentLocation
+	}
+	return location
 }
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -31,7 +31,6 @@ var (
 	DefaultPublisher = &Publisher{
 		APIKey:    os.Getenv("WORKOS_API_KEY"),
 		Endpoint:  "https://api.workos.com/events",
-		Retries:   3,
 		QueueSize: 512,
 	}
 )
@@ -48,6 +47,8 @@ type Event struct {
 	OccurredAt time.Time              `json:"occurred_at"`
 	TargetName string                 `json:"target_name"`
 	TargetID   string                 `json:"target_id"`
+
+	indempotencyKey string
 }
 
 // SetAPIKey sets the WorkOS API key to use when using Publish.
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -31,9 +31,8 @@ import (
 )
 
 var (
-	// DefaultPublisher is the publisher used by SetAPIKey, Publish and Close
-	// functions.
-	DefaultPublisher = &Publisher{
+	// DefaultClient is the client used by SetAPIKey and Publish functions.
+	DefaultClient = &Client{
 		APIKey:   os.Getenv("WORKOS_API_KEY"),
 		Endpoint: "https://api.workos.com/events",
 	}
@@ -47,12 +46,12 @@ func init() {
 
 // SetAPIKey sets the WorkOS API key to use when using Publish.
 func SetAPIKey(k string) {
-	DefaultPublisher.APIKey = k
+	DefaultClient.APIKey = k
 }
 
 // Publish publishes the given event.
 func Publish(ctx context.Context, e Event) error {
-	return DefaultPublisher.Publish(ctx, e)
+	return DefaultClient.Publish(ctx, e)
 }
 
 // Event represents an Audit Log event.
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -79,3 +79,10 @@ func defaultLocation(location string) string {
 	}
 	return location
 }
+
+func defaultTime(t time.Time) time.Time {
+	if t == (time.Time{}) {
+		t = time.Now()
+	}
+	return t
+}
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -70,7 +70,7 @@ type Event struct {
 	TargetName string                 `json:"target_name"`
 	TargetID   string                 `json:"target_id"`
 
-	indempotencyKey string
+	idempotencyKey string
 }
 
 func defaultLocation(location string) string {
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -1,0 +1,48 @@
+// Package alog is a package to send audit logs events to WorkOS.
+package alog
+
+import (
+	"os"
+	"time"
+)
+
+var (
+	// DefaultPublisher is the publisher used by SetAPIKey, Publish and Close
+	// functions.
+	DefaultPublisher = &Publisher{
+		APIKey:    os.Getenv("WORKOS_API_KEY"),
+		Endpoint:  "https://api.workos.com/events",
+		Retries:   3,
+		QueueSize: 512,
+	}
+)
+
+// Event represents an Audit Log event.
+type Event struct {
+	Action     string                 `json:"action"`
+	ActionType string                 `json:"action_type"`
+	ActorName  string                 `json:"actor_name"`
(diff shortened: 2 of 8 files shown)


pkg/auditlog/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How it works
+
+```go
+package main
+
+import "github.com/workos-inc/workos-go/pkg/auditlog"
+
+func main() {
+    alog.SetAPIKey("my_api_key")
```
maxchehab: Why the shortname? Does `auditlog.*` work?

pkg/auditlog/publisher.go
```diff
@@ -0,0 +1,158 @@
...
+	if p.Log == nil {
+		p.Log = log.Printf
+	}
+
+	if p.QueueSize < 1 {
+		p.QueueSize = 512
+	}
+
+	p.queue = make(chan Event, p.QueueSize)
+	p.stop = make(chan struct{})
```
dewski: Should this just be a `bool`?

pkg/auditlog/README.md
dewski: Had a similar thought, I would expect the package name to match the last path of the import.

pkg/auditlog/auditlog.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Close stops publishings audit log events and releases allocated resources.
+// It waits for pending events to be sent before returning.
+func Close() {
+	DefaultPublisher.Close()
+}
+
+// Event represents an Audit Log event.
+type Event struct {
+	Action     string                 `json:"action"`
```
dewski: There is an `Action` type:

```suggestion
	Action     Action                 `json:"action"`
```

pkg/auditlog/auditlog.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Close stops publishings audit log events and releases allocated resources.
+// It waits for pending events to be sent before returning.
+func Close() {
+	DefaultPublisher.Close()
+}
+
+// Event represents an Audit Log event.
+type Event struct {
```
dewski: Can the existing `Event` struct not be used?

pkg/auditlog/publisher.go
```diff
@@ -0,0 +1,158 @@
...
+		select {
+		case <-ctx.Done():
+			p.stop <- struct{}{}
+			return
+
+		case e := <-p.queue:
+			// This is to capture e value in order to not have the same value
+			// passed in different goroutines.
+			event := e
+			event.indempotencyKey = uuid.New().String()
```
dewski: It seems like the idempotency key should be set when the event is created, otherwise if they have a reference to an event struct they can pass the same event twice.

https://github.com/workos-inc/workos-go/blob/8ca00a6ef4cda750c32cf86075932a89c1afcac8/auditlog/event.go#L78-L92

pkg/auditlog/README.md
maxence-charriere: For verbosity, using a package in go requires that each functions or structs are prefixed by `pkgname.`. For this reason we usually try to keep package name short.

I'm not really opinionated about this so if we really feels it an important tradeoff in readability, just let it rename it `auditlog`.

maxchehab: Ah, good point. But I think we should also value continuity between the Go and Node sdk. Would you mind reverting it to `auditlog` to preserve this?

pkg/auditlog/auditlog.go
```diff
@@ -0,0 +1,88 @@
...
+func defaultLocation(location string) string {
+	if location == "" {
+		location = currentLocation
+	}
+	return location
+}
+
+func defaultTime(t time.Time) time.Time {
+	if t == (time.Time{}) {
+		t = time.Now()
```
dewski: Should send UTC time.

```suggestion
		t = time.Now().UTC()
```

pkg/auditlog/publisher.go
```diff
@@ -0,0 +1,158 @@
...
+	// https://dashboard.workos.com/api-keys.
+	APIKey string
+
+	// The http.Client that is used to post audit log events to WorkOS. Defaults
+	// to http.Client.
+	Client *http.Client
+
+	// The endpoint used to request Workos. Defaults to
+	// https://api.workos.com/events.
+	Endpoint string
```
maxchehab: On the subject of continuity, in the node sdk, we reference this as `apiHostname` 

https://github.com/workos-inc/workos-js/blob/2dd79c10bb9ef3dd0bbba3c11b3239ac63e79e72/src/common/interfaces/workos-options.interface.ts#L1-L4

pkg/auditlog/publisher.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// The function used to encode in JSON. Defaults to json.Marshal.
+	JSONEncode func(v interface{}) ([]byte, error)
```
dewski: Do we know when a custom JSON encoder would be used?

pkg/auditlog/README.md
dewski: From the [go docs](https://blog.golang.org/package-names):

> By convention, the last element of the package path is the package name

pkg/auditlog/publisher.go
maxchehab: Should the publisher be a package level model? Why can't it belong higher in the library's hierarchy?

pkg/auditlog/README.md
maxence-charriere: reverted

pkg/auditlog/publisher.go
maxence-charriere: The struct{} is a struct type with zero elements. It is often used when no information is to be stored. It has the benefit of being 0-sized, so usually no memory is required to store a value of type struct{}

pkg/auditlog/publisher.go
maxence-charriere: Yep for performance reason. Go default json encoder is slower and allocates more than other 3rd party serializer. Allowing this to be set can open to a serious optimization for the user.

For example, back on my days in Segment, we implemented a JSON encoder that literally blow away the go standard json encoder in term of performance and resource footprint.

pkg/auditlog/publisher.go
maxence-charriere: The event is passed by value. Event if they keep a reference to the event they published,
The event that is passed to the channel is a copy of the one they used in the `Publish` call. The reference they have will always have an indempotency key empty.

pkg/auditlog/auditlog.go
maxence-charriere: Why do we would not want to use it?

Using it allows to have a similar behavior as the JS SDK, it also self document how an event can be filled (it would be even better if we can add godoc above each field, explaining what it is and give directions on how to fill the values).

pkg/auditlog/auditlog.go
maxence-charriere: looks like the `Action` type in the current implementation provide methods to extract `Category`, `Name` and `Environment` from a string.

I was not able to find where those method are used in the implementation except in the `action_test.go`.

What would be the scenario where a user would use the type?

pkg/auditlog/publisher.go
maxence-charriere: The publisher is meant to continuously send events. I don't think this behavior would fit to every create/update scenario. If in the future we have other use cases that could benefit from this kind of logic, we would refactor in order to reuse code.

maxence-charriere: For the `apiHostname` if we have multiple `path` in order to deal with events. I the current state, the publisher in only dealing with 1 endpoint. 

Having it set once allow to a avoid the intermediary step of building the path like 
```go
endpoint := p.APIHostname + "/events"
```
at each request we send.

maxence-charriere: But I think this should be done that way you suggest for when the default publisher is initialized.

pkg/auditlog/auditlog.go
maxence-charriere: Sorry I might have misunderstood this one. The reason I did not want to reuse the current implementation is because:
- For now the action of this implementation does not use an Action type
- The current event implementation contains a logic to publish the event. People tend to take a look on godoc to learn how a package can be use, having 2 different way to do the same thing is confusing.

pkg/auditlog/auditlog.go
dewski: When this was implemented the API allowed you to prefix events with an environment as the first part of an action name separated by a dot. Working with action names I often try to avoid splitting the dots as much as I can, so this type encapsulates the logic so it doesn't need to be written every time you're trying to get the first part of the action name or the last part.

pkg/auditlog/auditlog.go
dewski: When this gets merged it wouldn't make sense to have the `auditlog` directory that currently exists. We could still carry over the convenience methods that exist on the `Event` struct while still publishing new events using this new approach.

For example, with the existing `Event` struct, other structs in an application can implement the `auditlog.Auditable` interface which makes it easy for applications to set the group, actor, and target for an event. From experience trying to set the IDs and human strings manually each time an action is logged is hard to maintain good data quality.

pkg/auditlog/publisher.go
dewski: > The event that is passed to the channel is a copy of the one they used in the `Publish` call.

Since the idempotency key isn't set until the channel receives it in this implementation, if the client sends the same struct again a new idempotency key will be generated for the same event.

pkg/auditlog/publisher.go
dewski: Should workos-go provide a better encoder by default if that is the case?

pkg/auditlog/auditlog.go
```diff
@@ -0,0 +1,88 @@
...
+	ActorName  string                 `json:"actor_name"`
+	ActorID    string                 `json:"actor_id"`
+	Group      string                 `json:"group"`
+	Location   string                 `json:"location"`
+	Metadata   map[string]interface{} `json:"metadata,omitempty"`
+	OccurredAt time.Time              `json:"occurred_at"`
+	TargetName string                 `json:"target_name"`
+	TargetID   string                 `json:"target_id"`
+
+	indempotencyKey string
```
dewski:
```suggestion
	idempotencyKey string
```

maxence-charriere: good catch

pkg/auditlog/publisher.go
maxence-charriere: I think going with the standard library is a solid choice. We should not enforce the use of a 3rd party library.

pkg/auditlog/auditlog.go
maxence-charriere: What I mean is that the sdk only provide create operation. Once an event is created, what is the scenario where I would need to parse the event string?

Also do we need a type for this?
I think a function with the following signature:
```go
func ParseEvent(event string) (category, name, env string) 
```
would be more adapted to the need of getting the different part of the event.

The difference come in term of usage, each time we want to fill an event, we have to wrap the value like `auditlog.Event("my.event.org")`.
Its pretty verbose just to handle the scenario where we would want to deconstruct it after publish.

dewski: The library only provides the create API now, but it will eventually provide a read API and query API. By also requiring a custom type it gives us the ability to call functions on it like `action.IsValid()` to ensure it is valid.

@maxchehab are there any validations on action names?

maxence-charriere: @dewski I updated my previous comment, I was referring to action.

pkg/auditlog/auditlog.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Event represents an Audit Log event.
+type Event struct {
+	Action     string                 `json:"action"`
+	ActionType string                 `json:"action_type"`
+	ActorName  string                 `json:"actor_name"`
+	ActorID    string                 `json:"actor_id"`
+	Group      string                 `json:"group"`
+	Location   string                 `json:"location"`
+	Metadata   map[string]interface{} `json:"metadata,omitempty"`
```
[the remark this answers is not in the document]
maxence-charriere: I forgot to implement the global metadata, I ll add it to the implementation.
I'm curious about the metadata validation.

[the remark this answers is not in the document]
maxence-charriere: ```One thing I would have added if I took another stab at it would be making the metadata field not exported and only mutable by exported functions. This would give us more control of the types of values the client would know they could attach to it.```

It could be a solution. Is it possible to have multiple level in metadata?
```go
meta := map[string]interface{}{
    "simple": 42,
    "nested": map[string]interface{}{
        "nested simple": 42,
    },
    "struct": struct{
        Field string
    }{
        Field: "hello", 
   },
}
```

pkg/auditlog/auditlog.go
maxence-charriere: Does the action format under the form of `category.action.env` is a good practice or it have to be enforced?

pkg/auditlog/auditlog.go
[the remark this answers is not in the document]
dewski: > I'm curious about the metadata validation.

The current implementation handles the API specs last I saw them, but @marktran and @maxchehab can chime in further if those are still accurate.

> It could be a solution. Is it possible to have multiple level in metadata?

I believe the answer is no, but @marktran and @maxchehab answer with more clarity.

pkg/auditlog/publisher.go
dewski: > It feels like reusing a same event is a developer mistake. Each event should be unique whether they embeds the same data.

For sure! I see the benefits of both implementations. I think clients should help guard against developer mistakes where it can. It will boil down to the semantics the client wants to enforce here, is the client treating the requests as idempotent or the events? If events are idempotent, the requests will be as a result. Just my thoughts on it.

pkg/auditlog/auditlog.go
maxence-charriere: The `Auditable` interface implies that the user implements the methods to their types to satisfy it.

Implementing the interface adds the work of adapting user models in order to satisfy our interface. Since they still must have to know what to return when implementing it, I feel it ends to the same result but with additional steps.

Automating how an event would be filled could also be done in a function that the user would implement following his personal needs without having to add a vendor logic to his data model.

Having 2 ways to achieve a same goal can be confusing.
When I'm reading the godoc, if I see that there is method to fill the Event, I will think this is the preferred way to use the package. It gonna lead me to implement something more heavy in order to go to the goal I seek.

pkg/auditlog/auditlog.go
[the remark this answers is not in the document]
maxence-charriere: Another question:
In the current global metadata implementation, in case of duplicate key, look like the event metadata will always be overridden by the global one:
```go
	// Add the global metadata to the Event's metadata
	for k, v := range globalMetadata {
		e.Metadata[k] = v
	}
```

Is it the behavior we want?
The one set at the event level are the one that are the one who are the most up to date. Have the feeling that they are the one that should prevail.

pkg/auditlog/publisher.go
maxence-charriere: Since it is documented in the the API, is the idempotency key should be a choice for the user?
We could just make this field exported and fill it only if it is not set by the user.

What I'm trying to keep is a declaration by structure since each field document by itself the Event content.

I would also like to keep passing an event by value in order to prevent a memory allocation on the stack.

pkg/auditlog/auditlog.go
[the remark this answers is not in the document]
dewski: An additional check can be performed here to see if the key is already set or the local metadata can override the global metadata. It's definitely a good question with no easy answer.

pkg/auditlog/auditlog.go
dewski: > Implementing the interface adds the work of adapting user models in order to satisfy our interface. Since they still must have to know what to return when implementing it, I feel it ends to the same result but with additional steps.

The `Auditable` interface is optional, but not required. The integrator can set `ActorName`, `ActorID`, `TargetName` and `TargetID` manually if they like.

> Automating how an event would be filled could also be done in a function that the user would implement following his personal needs without having to add a vendor logic to his data model.

Totally! Integrators can write their own functions should they choose.

> Having 2 ways to achieve a same goal can be confusing.

The interface is used for the convenience functions `Event.SetActor`, `Event.SetTarget`, and if it's ever a value in the metadata. From experience it's difficult to make sure the same value is logged for the same model across many engineering teams creating and maintaining Audit Log events.

For example, without an interface, if my app had events like `team.create`, `team.update`, `team.destroy`, `team.add_member`, `team.remove_member`, `repository.update_team_permission`, `team_discussion.create`, etc. and I wanted to attach the team name and team ID for each Audit Log event it'd have to be:

```go
event.Action = "team.create"
event.Metadata["team"] = team.NameWithOwner # => org-name/team-name
event.Metadata["team_id"] = team.ID # => 1
```

Repeating this for every event that wants to attach the team is repetitive and error prone. If another team adds a `team_discussion.destroy` event later down the road and they populate the team manually with the wrong function or field like so:

```go
event.Action = "team.create"
event.Metadata["team"] = team.Name # => team-name
event.Metadata["team_id"] = team.ID # => 1
```

It may not be caught in code review and the team name logged across all Audit Log actions will be inconsistent and could confuse administrators. The interface gives teams the opportunity to encapsulate the logic necessary in one place and make it testable.

```go
func (t Team) ToAuditableName() string {
  return fmt.Sprintf("%s/%s", t.OrganizationName, t.Name)
}

func (t Team) ToAuditableID() string {
  return strconv.Itoa(t.DatabaseID)
}

event.Action = "team.create"
event.Metadata["team"] = team
```

Which will result in:

```json
{
  "team": "org-name/team-name",
  "team_id": "1"
}
```

> I think the first step to go toward data quality is providing good documentation. Adding godoc above each fields explaining what it is and how to fill it could greatly contribute to data quality.

💯 we should definitely add documentation for each field when possible.

pkg/auditlog/auditlog.go
maxchehab: > are there any validations on action names?
@dewski 

Nope, any names are acceptable.

pkg/auditlog/auditlog.go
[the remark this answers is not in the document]
marktran: @maxence-charriere We talked about having a metadata limit similar to Stripe:

> You can specify up to 50 keys, with key names up to 40 characters long and values up to 500 characters long.
\- https://stripe.com/docs/api/metadata

This isn't enforced consistently at the moment.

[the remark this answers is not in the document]
maxence-charriere: This could be documented. I did not see that those check are done in the js sdk. do we want to throw an error if a key is longer than 50 character or do we want to trim the key? How this is handled in js?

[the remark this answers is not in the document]
marktran: Added a story internally for this.

maxence-charriere:
Ok folks, I refactored this PR taking care of some feedback:
- Global metadata are back
- Metadata amount and keys are checked:
  - Max 50
  - Key max length 40
- When merging metadata and global metadata, global metadata does not override event metadata anymore
- Queuing/background publishing logic has been removed. The reason for this is:
  - Consistency with the JS SDK
  - Handling concurrent publishing should not be enforced by this package, we should let the integrators decide what is the best solution to fit their needs related to concurrency or the overhead of sending an audit log

Things that I did not add back:
- Setter: The setter/auditable really confused me when I used this SDK for the first time. Also one of the feedback from the private beta was why are we using setter with the go library. 

Please give another review on this.

pkg/auditlog/auditlog.go
maxchehab: I agree with @maxence-charriere in regards to simplifying the sdk.

These setters add additional implementation requirements for consumers. They  will have to redesign all of their current data models in-order to use this interface.

This decision seems to only impact larger organizations where multiple teams have difficulty synchronizing a Target, Action, or Actor name.

I think our out-of-the box solution should not force a small organization, with no synchronization problems to sit in the same boat as a large org.

maxchehab:
This LGTM @maxence-charriere :shipit:

pkg/auditlog/auditlog.go
dewski: > These setters add additional implementation requirements for consumers. They will have to redesign all of their current data models in-order to use this interface.

The setters are optional, not required. Same with the `Auditable` interface. The `Auditable` interface isn't exclusive for only the actor and target fields, it is also applicable for any struct that gets included in the metadata. That could be up to 50 properties.

I haven't suggested that teams are required to use the `SetActor` or `SetTarget` functions, I'm mostly trying to highlight what the SDK and eventual data would look like without an interface that small to large companies _should_ implement.

The interface lets the SDK know how each struct should be represented for the human readable string and for the machine readable identifier. Teams implementing the SDK could still construct an `Event` struct manually and set each field as demonstrated in this PR. Once someone sets up 10 unique events in the Audit Log they will inevitably start sharing structs across events, and you risk data quality issues such as mixed key names representing the same value across events (for example `org` in `team.remove_member` and `organization` in `team.add_member`), inconsistent values logged for the same property (for example for a team it would be `workos-inc/employees` in one event and just `employees` in another).

For example, on a `team.add_member` event with no metadata we either have to set the target as the user being added or the team the user is being added to:

```json
{
  "actor": "dewski",
  "actor_id": "User;79995",
  "target": "maxchehab",
  "target_id": "User;12345",
  "metadata": {}
}
```

or:

```json
{
  "actor": "dewski",
  "actor_id": "User;79995",
  "target": "workos-inc/employees",
  "target_id": "Team;12345",
  "metadata": {}
}
```

To start with we'll at least have to add the `team` or `user` to metadata to get a clearer picture of what happened, so I'll just stick with the target being the user being added as the base event:

```json
{
  "actor": "dewski",
  "actor_id": "User;79995",
  "target": "maxchehab",
  "target_id": "User;12345",
  "metadata": {
    "team": "workos-inc/employees",
    "team_id": "Team;12345"
  }
}
```

If we need to know the organization that the team belonged to at the time of the event we'll need to add that:

```json
{
  "actor": "dewski",
  "actor_id": "User;79995",
  "target": "maxchehab",
  "target_id": "User;12345",
  "metadata": {
    "team": "workos-inc/employees",
    "team_id": "Team;12345",
    "organization": "workos-inc",
    "organization_id": "Organization;12345",
    "business": "workos-inc",
    "business_id": "Business;12345"
  }
}
```

As is the Go struct would look something like this:

```go
event := auditlog.Event{
  Action: "team.add_member",
  Actor: actor.Login,
  ActorID: actor.ID,
  Target: user.Login,
  TargetID: user.ID,
  Metadata: map[string]interface{}{
    "team": fmt.Sprintf("%s/%s", organization.Login, team.Slug),
    "team_id": team.ID,
    "organization": organization.Login,
    "organization_id": organization.ID,
    "business": business.Login,
    "business_id": business.ID,
  }
}
```

If someone wanted to add a `team.remove_member` event a few months later they'll either have to know to copy and paste this struct as a starting point, or build it out by hand and may not know the exact values to be used. If every time someone is expected to add an event, values will start to drift since there isn't anything telling them what they should be.

```go
event := auditlog.Event{
  Action: "team.remove_member",
  Actor: actor.Login,
  ActorID: actor.ID,
  Target: user.Login,
  TargetID: user.ID,
  Metadata: map[string]interface{}{
    "team": team.Name,
    "teamId": team.ID,
    "org": organization.Name,
    "orgId": organization.ID,
    "business": business.Name,
    "businessId": business.ID,
  }
}
```

_Take note of the inconsistent key names, `team`/`org`/`business` values._

If I'm to ever consume this data pragmatically I'll have to always account for the organization being stored in an `org` and `orgId` field. While you can't _prevent_ it from happening on purpose, I believe the SDK should do its best to try and prevent that from happening.

My only suggestion has been to encourage the use of interfaces to prevent drifting data, consistent key names, and values. The alternative way to write these events would be:

```go
event := auditlog.Event{
  Action: "team.remove_member",
  Actor: actor.Login,
  ActorID: actor.ID,
  Target: user.Login,
  TargetID: user.ID,
  Metadata: map[string]interface{}{
    "team": team,
    "organization": organization,
    "business": business,
  }
}
```

It makes it less ambiguous what the values should be and reduce the chance of data quality issues. While large teams could create their own interface and functions to generate events, it's putting a lot of work on each company integrating with WorkOS to roll it by hand.

I think the SDK should provide helper utilities such as the interface to get the most out of the tool. I didn't touch into other fields that you'd want to log across each event like `request_id`, `http_method`, `http_path`, etc. If I had to set those manually each time I added a new event it'd require more work on my part and more room for data quality issues where they could be missing completely and make it difficult to debug issues in the future.

> I think our out-of-the box solution should not force a small organization, with no synchronization problems, to sit in the same boat as a large org.

Small teams will eventually run into this as they start to grow and get more out of the WorkOS Audit Log.

PR #16, opened 2019-10-29
Simplify sso usage

maxence-charriere:
## Summary
- `ProjectID` is now set at the client level
- `RedirectURI` is now set at the client level
- `DefaultClient` is now initialized with `Configure()` function

## Why this change
I noticed that both `ProjectID` and `RedirectURI` are used in `GetAuthorizationURL` and `GetProfile`:
- It is required that they match between those 2 calls
- Having them set once is less prone to human error than documenting that they must match
- Less options to think about => more simple => better dev experience when getting started

### DIFF
--- a/pkg/sso/README.md
+++ b/pkg/sso/README.md
@@ -16,18 +16,16 @@ import (
 )
 
 func main() {
-    sso.SetAPIKey("my_api_key")
+    sso.Init("my_api_key", "my_workos_project_id")
 
     http.Handle("/login", sso.Login(sso.GetAuthorizationURLOptions{
         Domain:      "mydomain.com",
-        ProjectID:   "my_workos_project_id",
         RedirectURI: "https://mydomain.com/callback",
     }))
 
     http.HandleFunc("/callback", func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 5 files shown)


pkg/sso/README.md
```diff
@@ -16,19 +16,19 @@ import (
...
 
 func main() {
-    sso.SetAPIKey("my_api_key")
+    sso.Configure(
+        "xxxxx",                            // WorkOS api key
+        "project_xxxxx",                    // WorkOS project id
+        "https://mydomain.com/callback",    // Redirect URI
+    )
 
     http.Handle("/login", sso.Login(sso.GetAuthorizationURLOptions{
```
maxchehab:
```suggestion
    http.Handle("/login", sso.Login(sso.GetAuthorizationURLOptions {
```

Maybe run this code thru the formater?

maxence-charriere: Formatter does not work in the readme. But my bad, I have been lazy.

maxence-charriere: Actually I just double checked it on vscode, seems alright

```go
sso.GetAuthorizationURLOptions{
```
no space on this one after options

PR #22, opened 2019-12-27
Maxchehab/ch1167/remove redirect uri from the getprofile function

maxchehab:
This PR:
- removes the `redirect_uri` query parameter from the `POST /sso/token` request.
- ~adds a [**Release Notes**](https://github.com/workos-inc/workos-go/tree/maxchehab/ch1167/remove-redirect-uri-from-the-getprofile-function#release-notes) section to the readme~


jthodge:
I like @marktran's [idea](https://github.com/workos-inc/workos-js/pull/49#discussion_r361723806) of keeping a separate `CHANGELOG.md` file instead of release notes living in `README.md`.

Think we should do the same thing here for consistency?

PR #24, opened 2020-01-22
Maxchehab/ch1327/go implement idempotency key header for audit

maxchehab:
This PR adjusts the idempotency key feature of the SDK so that if it is not provided by the developer, the header will not be attached to the outgoing request.

I had to remove the idempotency key as a required header within the Audit Log tests (which lowers coverage) since I was unsure how to test this change without making drastic changes to how the test suite functions.

### DIFF
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -67,7 +67,7 @@ type Event struct {
 
 	// If no key is provided or the key is empty, the key will not be attached
 	// to the request.
-	IdempotencyKey string
+	IdempotencyKey string `json:"-"`
 
 	// An ip address that locates where the audit log occurred.
 	Location string `json:"location"`
--- a/pkg/auditlog/auditlog.go
+++ b/pkg/auditlog/auditlog.go
@@ -28,8 +28,6 @@ import (
 	"fmt"
 	"os"
 	"time"
-
-	"github.com/google/uuid"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


marktran:
[ch1327]

Do we need to connect CH?

pkg/auditlog/auditlog.go
```diff
@@ -63,12 +61,13 @@ type Event struct {
...
+	// A key that ensures that the same event is not processed multiple times.
 	// Once the event is sent for the first time, the lock on the key expires
 	// after 24 hours.
-	//
-	// An idempotency key is automatically generated if not set.
-	IdempotencyKey string `json:"-"`
+
+	// If no key is provided or the key is empty, the key will not be attached
+	// to the request.
+	IdempotencyKey string
```
maxence-charriere: This will include the idempotency key in the JSON payload under the `IdempotencyKey`.
Since we are including it in the header I don't think we should remove the `json:"-"`

pkg/auditlog/client.go
```diff
@@ -67,9 +66,12 @@ func (c *Client) Publish(ctx context.Context, e Event) error {
 	}
 	req = req.WithContext(ctx)
 	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("Idempotency-Key", e.IdempotencyKey)
 	req.Header.Set("Authorization", "Bearer "+c.APIKey)
 
+	if len(e.IdempotencyKey) > 0 {
```
maxence-charriere: little style stuff:
use 
```go
e.IdempotencyKey != ""
```
to compare string. The compiler does the optimization by himself and when you read the code you know directly that we deal with a string.

Use the `len(xx) != 0` when you check a slice or map length.

maxchehab: That's good to know!

sbauch:
so as soon as this is merged its "published" right? theres no go package repository?

but typically a Go developer will "go get" a package once and it saves the source to their codebase?  I guess im just unclear how we handle versioning here? can chat offline

maxence-charriere:
@sbauch versioning is handled by `go get` and take a repo release tag.
All we need is tag a new release in github.

PR #25, opened 2020-01-30
Allow for optional provider parameter in authorization URL

rjvani:


pkg/sso/client.go
```diff
@@ -87,11 +91,18 @@ func (c *Client) GetAuthorizationURL(opts GetAuthorizationURLOptions) (*url.URL,
 	c.once.Do(c.init)
 
 	query := make(url.Values, 5)
-	query.Set("domain", opts.Domain)
 	query.Set("client_id", c.ProjectID)
 	query.Set("redirect_uri", c.RedirectURI)
 	query.Set("response_type", "code")
 
+	if opts.Domain == "" && opts.Provider == "" {
+		return nil, errors.New("Incomplete arguments. Need to specify either a 'domain' or 'provider'")
```
maxence-charriere: Should not use capital letters in error messages.
=> https://github.com/golang/go/wiki/Errors

```
errors.New("incomplete arguments: missing domain or provider")
```

pkg/sso/client.go
```diff
@@ -87,11 +91,18 @@ func (c *Client) GetAuthorizationURL(opts GetAuthorizationURLOptions) (*url.URL,
...
+	} else if opts.Domain == "" {
+		query.Set("provider", opts.Provider)
```
maxence-charriere: is it possible to have `domain` and `provider` set at the same time?

if not we should do this check and throw an error.

sbauch:
Note that the API expects `"GoogleOAuth"`. seems this is just going to take an arbitrary string for provider?

maxence-charriere:
@sbauch what are the possible values for provider?

sbauch:
@maxence-charriere I _think_ just "`GoogleOAuth`" is technically valid )all the other providers require domain). but this is the enum: 

https://github.com/workos-inc/workos/blob/master/packages/api/src/common/enums/connection-type.enum.ts#L3

maxence-charriere:
@rjvani would be worth adding a type for provider =>

```go
// Provider represents a SSO provrider.
type Provider string

// Constans that enumerates SSO providers.
const (
    GoogleOAuth Provider = "GoogleOAuth"
)
```

Then replace the string by Provider in the options.

pkg/sso/client.go
```diff
@@ -74,6 +75,9 @@ type GetAuthorizationURLOptions struct {
 	// The app/company domain without without protocol (eg. example.com).
 	Domain string
 
+	// Authentication service provider descriptor.
```
maxence-charriere: We should add in the documentation that this is currently used only for Google as a provider.

pkg/sso/client.go
maxence-charriere: As you pointed on Slack, the API is doing a check on the provider first.
Doing a check on the provider first would improve code readability
```go
else if opts.Provider != "" {
    query.Set("provider", opts.Provider)
} else {
    query.Set("domain", opts.Domain)
}
```

pkg/sso/client.go
```diff
@@ -74,6 +77,10 @@ type GetAuthorizationURLOptions struct {
...
+	// Currently only used for Google.
```
maxence-charriere: He specify the `GoogleOAuth` value

pkg/sso/client_test.go
```diff
@@ -50,6 +59,30 @@ func TestClientAuthorizeURL(t *testing.T) {
 	}
 }
 
+func TestClientAuthorizeURLWithNoDomainAndProvider(t *testing.T) {
+	test := struct {
```
maxence-charriere: Look like you don't need this structure since you are not running some other sub tests.
You can inline the options:
```go
u, err := client.GetAuthorizationURL(GetAuthorizationURLOptions{
    State: "state",
})
```

pkg/sso/client.go
```diff
@@ -87,11 +94,18 @@ func (c *Client) GetAuthorizationURL(opts GetAuthorizationURLOptions) (*url.URL,
 	c.once.Do(c.init)
 
 	query := make(url.Values, 5)
-	query.Set("domain", opts.Domain)
 	query.Set("client_id", c.ProjectID)
 	query.Set("redirect_uri", c.RedirectURI)
 	query.Set("response_type", "code")
 
+	if opts.Domain == "" && opts.Provider == "" {
+		return nil, errors.New("incomplete arguments. Need to specify either a 'domain' or 'provider'")
```
maxence-charriere: see my previous comment on error messages

maxence-charriere: `incomplete arguments: missing domain or provider`

It avoids capital letters and we don't need to suggest what to do since the error message is pretty clear about the issue.

rjvani: Fixed

pkg/sso/client_test.go
```diff
@@ -50,6 +59,22 @@ func TestClientAuthorizeURL(t *testing.T) {
...
+		RedirectURI: "https://example.com/sso/workos/callback",
+	}
+
+	u, err := client.GetAuthorizationURL(GetAuthorizationURLOptions{
+		State: "state",
+	})
+
+	var expectedURL *url.URL = nil
+	require.Error(t, err)
+	require.Equal(t, expectedURL, u)
```
maxence-charriere: here you can delete 
```go
var expectedURL *url.URL = nil
```
and use
```go
require.Nil(t, u)
```
instead.

pkg/sso/client_test.go
```diff
@@ -33,6 +34,14 @@ func TestClientAuthorizeURL(t *testing.T) {
 			},
 			expected: "https://api.workos.com/sso/authorize?client_id=proj_123&domain=lyft.com&redirect_uri=https%3A%2F%2Fexample.com%2Fsso%2Fworkos%2Fcallback&response_type=code&state=custom+state",
 		},
+		{
+			scenario: "generate url with state",
```
maxence-charriere: Should rename this scenario by `generate url with provider`.

rjvani: 👍

pkg/sso/client_test.go
rjvani: 👍

pkg/sso/client.go
```diff
@@ -74,6 +77,10 @@ type GetAuthorizationURLOptions struct {
 	// The app/company domain without without protocol (eg. example.com).
 	Domain string
 
+	// Authentication service provider descriptor.
+	// Currently only used for ConnectionType.GoogleOAuth
```
maxence-charriere: We are not accessing this with `ConnectionType.GoogleOAuth`. It may mislead the reader.

=> `Provider is currently only used when the connection type is GoogleOAuth.`

sbauch:
can we merge both open PRs and bump the version then? thanks so much for taking care of this work gang, nobody on dev ex knows go 🙈

rjvani:
> can we merge both open PRs and bump the version then? thanks so much for taking care of this work gang, nobody on dev ex knows go 🙈

Definitely 👍 I need a final look on the Python one, but can merge that today as well.

PR #26, opened 2020-02-05
Release documentation

maxence-charriere:
## Summary
- Use `go mod` to define package dependencies
- The client uses a centralized version number in the user agent
- Makefile that helps release a version number
  ```sh
   make VERSION=v0.0.0 release
   ```
- Update README with release section


makefile
```diff
@@ -0,0 +1,18 @@
+test:
```
rjvani: Nit: I think these files are usually named `Makefile` by default

maxence-charriere: I renamed it but look like GitHub keep it with the lowercase.

PR #30, opened 2020-02-14
Add both domain and provider when building SSO authorization URL

rjvani:


pkg/sso/client.go
```diff
@@ -100,9 +100,11 @@ func (c *Client) GetAuthorizationURL(opts GetAuthorizationURLOptions) (*url.URL,
 
 	if opts.Domain == "" && opts.Provider == "" {
 		return nil, errors.New("incomplete arguments: missing domain or provider")
-	} else if opts.Provider != "" {
+	}
+	if opts.Provider != "" {
 		query.Set("provider", string(opts.Provider))
```
maxchehab: do we need to cast `opts.Provider` as a string?

rjvani: Yeah, the compiler complains otherwise: `cannot use opts.Provider (type ConnectionType) as type string in argument to query.Set`.

maxchehab: Ah, I missed that we defined that as a type.

PR #36, opened 2020-03-02
rename ID to Token

marktran:
Rename ID to Token.


pkg/sso/client.go
```diff
@@ -198,7 +198,7 @@ func (c *Client) GetProfile(ctx context.Context, opts GetProfileOptions) (Profil
 // PromoteDraftConnectionOptions contains the options to pass in order to
 // promote a draft connection.
 type PromoteDraftConnectionOptions struct {
-	ID string `json:"id"`
+	Token string `json:"id"`
```
rjvani: Should we rename the key as well?

marktran: I actually know how this works. @maxchehab ?

PR #41, opened 2020-04-16
Create package for Directory Sync

rjvani:
Offers functions to retrieve Directory data via the WorkOS API.

Retrieve Directory Users
```go
directorysync.ListUsers(context.Background(), ListUsersOpts{
	Directory: "directory_id",
})
```

Retrieve Directory Groups
```go
directorysync.ListGroups(
	context.Background(),
	ListGroupsOpts{
		Directory: "directory_id",
	},
)
```

Retrieve a single Directory User
```go
directorysync.GetUser(context.Background(), GetUserOpts{
	User:     "directory_usr_id",
})
```

Retrieve a single Directory Group
```go
directorysync.GetGroup(
	context.Background(),
	GetGroupOpts{
		Group:     "directory_grp_id",
	},
)
```

Retrieve a Project's Directories
```go
directorysync.ListDirectories(
	context.Background(),
	ListDirectoriesOpts{},
)
```

### DIFF
--- a/pkg/directorysync/client.go
+++ b/pkg/directorysync/client.go
@@ -78,7 +78,7 @@ type ListMetadata struct {
 	// Pagination cursor to receive records before a provided ID.
 	Before string `json:"before"`
 
-	// Pagination cursor to receive records before a provided ID.
+	// Pagination cursor to receive records after a provided ID.
 	After string `json:"after"`
 }
 
@@ -316,7 +316,6 @@ func (c *Client) GetGroup(
 ) (Group, error) {
 	c.once.Do(c.init)
 
-	fmt.Printf("%+v\n", opts)
 	endpoint := fmt.Sprintf(
 		"%s/directory_groups/%s",
 		c.Endpoint,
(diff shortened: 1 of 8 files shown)


pkg/directorysync/client.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+	defer res.Body.Close()
+
+	if err = workos.TryGetHTTPError(res); err != nil {
+		return ListUsersResponse{}, err
+	}
+
+	var body ListUsersResponse
+	dec := json.NewDecoder(res.Body)
+	err = dec.Decode(&body)
```
maxchehab, line 163: Can we default to a user provided JSON encoder like the SSO or Audit Trail packages.

https://github.com/workos-inc/workos-go/blob/master/pkg/sso/client.go#L56-L57

rjvani: Hmm, do we need a JSON encoder? The requests are all `GET`, and we're only passing query parameters for the resource IDs or pagination options.

I don't believe it's being used in SSO either [here](https://github.com/workos-inc/workos-go/blob/master/pkg/sso/client.go#L157).

maxchehab: Ah, my bad, we are just decoding stuff here 👍

PR #42, opened 2020-05-01
Add CreateConnection to SSO package

rjvani:


pkg/sso/client.go
```diff
@@ -227,3 +231,104 @@ func (c *Client) PromoteDraftConnection(ctx context.Context, opts PromoteDraftCo
...
+	OAuthSecret string `json:"oauth_secret"`
+
+	// OAuth Client Redirect URI.
+	OAuthRedirectURI string `json:"oauth_redirect_uri"`
+
+	// Identity Provider Issuer.
+	SamlEntityID string `json:"saml_entity_id"`
+
+	// Identity Provider SSO URL.
+	SamlIdpURL string `json:"saml_idp_url"`
```
maxchehab, line 285:
```suggestion
	SamlIDPURL string `json:"saml_idp_url"`
```
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

rjvani: Ah good catch. Updated.

PR #53, opened 2020-09-16
Feature/ch3304/add list organizations method to the go sdk

maxchehab:
Adds a `portal` package and `ListOrganizations` method to interface with our Admin Portal API.

### DIFF
--- a/pkg/portal/client.go
+++ b/pkg/portal/client.go
@@ -5,10 +5,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strconv"
 	"sync"
 	"time"
 
+	"github.com/google/go-querystring/query"
 	"github.com/workos-inc/workos-go/internal/workos"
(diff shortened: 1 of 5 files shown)


pkg/portal/client.go
```diff
@@ -0,0 +1,142 @@
...
+
+	limit := ResponseLimit
+	if opts.Limit != 0 {
+		limit = opts.Limit
+	}
+	q := req.URL.Query()
+
+	if len(opts.Domains) > 0 {
+		for _, domain := range opts.Domains {
+			q.Add("domains", domain)
```
rjvani, line 120: We're using [`go-querystring`](https://github.com/google/go-querystring) when listing Audit Trail events [here](https://github.com/workos-inc/workos-go/blob/master/pkg/audittrail/client.go#L142). Could we do something similar here?

rjvani (approved):

PR #56, opened 2020-10-13
Expose `connection_id` on Profile Objects

jthodge:
* Passes the `connection_id` Profile object attribute through to the SDK. We currently expose `connection_id` via the `/sso/token` API endpoint.
* Updates the SDK version to `0.6.0`

### DIFF
--- a/internal/workos/workos.go
+++ b/internal/workos/workos.go
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v0.5.0"
+	Version = "v0.6.0"
 )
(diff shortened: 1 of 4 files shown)


internal/workos/workos.go
```diff
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v0.5.0"
+	Version = "v0.6.0"
```
maxchehab, line 5: This isn't a breaking change so I think we can just bump to 0.5.1

maxchehab (approved):

rjvani (approved):
Think there's a minor formatting change needed, but lgtm otherwise

internal/workos/workos.go
jthodge: @maxchehab @rjvani I think this would be minor bump based on our [versioning policy](https://github.com/workos-inc/workos/wiki/SDK-Release-Process#versioning-policy) since this isn’t a bug fix.

PR #62, opened 2020-11-23
Add RedirectURI Passwordless Option

jthodge:


pkg/passwordless/client.go
```diff
@@ -85,6 +85,10 @@ type CreateSessionOpts struct {
 	// REQUIRED
 	Type PasswordlessSessionType `json:"type"`
 
+	// Optional string value used to set the location
+	// that the user will be redirected to after authenticating
+	RedirectURI string `json:"redirectURI"`
```
rjvani, line 90: This should be `json:"redirect_uri"`

rjvani (changes requested):

pkg/passwordless/client.go
jthodge: Thanks, good eye!

PR #66, opened 2021-02-01
Decode Org ID from Connection API responses

Frezzle:
We want to get org ID from the GetConnection and ListConnection APIs, and they do return it, but the struct that the response is decoded into did not include it.


rjvani (approved):
Thanks!

PR #68, opened 2021-02-01
Add `DELETE /connections/:id`

jthodge:

- Adds functionality to delete a single Connection using the WorkOS SSO REST API

### DIFF
--- a/pkg/sso/sso.go
+++ b/pkg/sso/sso.go
@@ -78,6 +78,7 @@ func ListConnections(
 	return DefaultClient.ListConnections(ctx, opts)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/sso/sso.go
```diff
@@ -76,3 +77,10 @@ func ListConnections(
 ) (ListConnectionsResponse, error) {
 	return DefaultClient.ListConnections(ctx, opts)
 }
+
+func DeleteConnection(
```
rjvani, line 81: Nit: Can we add a comment for this like other functions?

rjvani (approved):

PR #74, opened 2021-02-08
Remove Unneeded Attribute from `Directory` Struct

jthodge:

* Removes `bearer_token` attribute from `Directory` struct


rohanjadvani (approved):
❤️

PR #86, opened 2021-04-22
Updating WorkOS go SDK to include delete directory function

sheldonvaughn:
PR to add a delete directory function into the WorkOS go SDK.


rohanjadvani:
Could we add a test case?

rohanjadvani (approved):

PR #104, opened 2021-06-01
Export Public Error Functions

cobbinma:
[issue](https://github.com/workos-inc/workos-go/issues/83)

pr to export a public error function that will allow me to tell whether a request failed to it being a bad request or not

this will save me writing code like:
```go
if strings.Contains(err.Error(), "400 Bad Request") {
   return nil, internal.ErrInvalidCode
}

### DIFF
--- a/pkg/workos_errors/errors.go
+++ b/pkg/workos_errors/errors.go
@@ -6,7 +6,7 @@ import (
 	"net/http"
 )
 
-func ErrIsBadRequest(err error) bool {
+func IsBadRequest(err error) bool {
 	var httpError workos.HTTPError
 	return errors.As(err, &httpError) && httpError.Code == http.StatusBadRequest
 }
--- a/pkg/workos_errors/errors.go
+++ b/pkg/workos_errors/errors.go
@@ -1,0 +1,12 @@
+package workos_errors
+
+import (
+	"errors"
+	"github.com/workos-inc/workos-go/internal/workos"
+	"net/http"
+)
+
+func ErrIsBadRequest(err error) bool {
+	var httpError workos.HTTPError
+	return errors.As(err, &httpError) && httpError.Code == http.StatusBadRequest
+}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/workos_errors/errors.go [hunk 1]
```diff
@@ -0,0 +1,12 @@
+package workos_errors
+
+import (
+	"errors"
+	"github.com/workos-inc/workos-go/internal/workos"
+	"net/http"
+)
+
+func ErrIsBadRequest(err error) bool {
```
maxdeviant, line 9: Could we change this name to `IsBadRequest`?

cobbinma: @maxdeviant done 👍

maxdeviant (approved):

pkg/workos_errors/errors_test.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func TestIsBadRequest(t *testing.T) {
+	type args struct {
+		err error
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
```
rohanjadvani, line 18: nit: We use `expected` instead of `want` in our test suite. Could we `s/want/expected` to be consistent?

pkg/workos_errors/errors_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+		{
+			name: "nil",
+			args: args{err: nil},
+			want: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := workos_errors.IsBadRequest(tt.args.err); got != tt.want {
```
rohanjadvani, line 47: Can we do:

```go
require.Equal(t, tt.expected, workos_errors.IsBadRequest(tt.args.err))
```

rohanjadvani (commented):
Some small nits to be consistent with the rest of the test suite.

PR #107, opened 2021-06-14
expose expires_in parameter when creating passwordless sessions

sheldonvaughn:
PR to expose expires_in, also added field to unit testing.


pkg/passwordless/client.go
```diff
@@ -61,6 +61,9 @@ type PasswordlessSession struct {
 	// ISO-8601 datetime at which the Passwordless Session link expires.
 	ExpiresAt string `json:"expires_at"`
 
+	//Remaining duration of a passwordless session link.
```
maxdeviant, line 64:
```suggestion
	// Remaining duration of a passwordless session link.
```

PR #111, opened 2021-07-07
add organization id to directory

cobbinma:
hey guys 👋 

pr to add `organization_id` field to `directory` as shown [here](https://workos.com/docs/reference/directory-sync/directory)

we'll be able to link directories to an organization using this


maxdeviant (approved):
Thanks for the PR!

PR #118, opened 2021-08-11
add created_at and updated_at fields

torigodfrey:
Resolves SDK-252.

Added created_at and updated_at to organization, connection, and directory.

### DIFF
--- a/pkg/directorysync/client.go
+++ b/pkg/directorysync/client.go
@@ -421,6 +421,12 @@ type Directory struct {
 
 	// Identifier for the Directory's Organization.
 	OrganizationID string `json:"organization_id"`
+
+	//Created At
+	CreatedAt string `json:"created_at"`
+
+	//Updated At
+	UpdatedAt string `json:"updated_at"`
 }
 
 // ListDirectoriesOpts contains the options to request a Project's Directories.
--- a/pkg/directorysync/client.go
+++ b/pkg/directorysync/client.go
@@ -422,10 +422,10 @@ type Directory struct {
 	// Identifier for the Directory's Organization.
 	OrganizationID string `json:"organization_id"`
 
-	//Created At
+	// The timestamp of when the Directory was created.
 	CreatedAt string `json:"created_at"`
 
-	//Updated At
+	// The timestamp of when the Directory was updated.
 	UpdatedAt string `json:"updated_at"`
 }
 
--- a/pkg/organizations/client.go
+++ b/pkg/organizations/client.go
@@ -68,6 +68,12 @@ type Organization struct {
 
 	// The Organization's Domains.
 	Domains []OrganizationDomain `json:"domains"`
+
+	//Created At
+	CreatedAt string `json:"created_at"`
+
+	//Updated At
+	UpdatedAt string `json:"updated_at"`
 }
 
 // GetOrganizationOpts contains the options to request details for an Organization.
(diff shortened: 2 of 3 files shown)


pkg/directorysync/client.go
```diff
@@ -421,6 +421,12 @@ type Directory struct {
 
 	// Identifier for the Directory's Organization.
 	OrganizationID string `json:"organization_id"`
+
+	//Created At
```
maxdeviant, line 425: Could we make these doc comments look more like this?

- Full sentences describing the field
- Space between the `//` and the first word of the comment

```suggestion
	// The timestamp of when the Directory was created.
```

torigodfrey: Done!

pkg/organizations/client.go
```diff
@@ -68,6 +68,12 @@ type Organization struct {
 
 	// The Organization's Domains.
 	Domains []OrganizationDomain `json:"domains"`
+
+	//Created At
```
mthadley, line 72: Similar to @maxdeviant's previous comment, we should update these comments (and the ones in `pkg/sso/client.go`).

After that, I think this will be good to go!

mthadley (changes requested):

pkg/organizations/client.go
torigodfrey: Hopefully got it all :)

mthadley: Looks like there's just a few more for `Organization` in `pkg/organizations/client.go`.

torigodfrey: Brutal that I missed that but this is now updated in: 

`pkg/sso/client.go`
`pkg/organizations/client.go`
`pkg/directorysync/client.go `

maxdeviant (approved):

pkg/organizations/client.go
```diff
@@ -68,6 +68,12 @@ type Organization struct {
 
 	// The Organization's Domains.
 	Domains []OrganizationDomain `json:"domains"`
+
+	// The timestamp of when the Directory was created.
```
mthadley, line 72:
```suggestion
	// The timestamp of when the Organization was created.
```

pkg/organizations/client.go
```diff
@@ -68,6 +68,12 @@ type Organization struct {
...
+	CreatedAt string `json:"created_at"`
+
+	// The timestamp of when the Directory was updated.
```
mthadley, line 75:
```suggestion
	// The timestamp of when the Organization was updated.
```

pkg/sso/client.go
```diff
@@ -321,6 +321,12 @@ type Connection struct {
 
 	// Domain records for the Connection.
 	Domains []ConnectionDomain `json:"domains"`
+
+	// The timestamp of when the Directory was created.
```
mthadley, line 325:
```suggestion
	// The timestamp of when the Connection was created.
```

pkg/sso/client.go
```diff
@@ -321,6 +321,12 @@ type Connection struct {
...
+	CreatedAt string `json:"created_at"`
+
+	// The timestamp of when the Directory was updated.
```
mthadley, line 328:
```suggestion
	// The timestamp of when the Connection was updated.
```

mthadley:
Just realized that some of the comments were referring to the wrong type. Left some suggestions that should fix it. Last thing, I think!

pkg/organizations/client.go
torigodfrey: 😅

mthadley (approved):
🎉

PR #127, opened 2021-11-12
Update name to email in PasswordlessSession

AshGodfrey:
Linear: 
https://linear.app/workos/issue/SUP-160/fix-passwordless-bug-in-go-sdk

flagged by @mthadley


mthadley (approved):
Thanks for fixing!

PR #129, opened 2021-11-17
export validate payload

cobbinma:
[Issue](https://github.com/workos-inc/workos-go/issues/128#issue-1056102581)

👋 pr to export validate payload in `webhooks`

also added a method to inject the current time dependency (for testing).


mthadley (approved):
Thanks for the contribution, @cobbinma!

After this is merged, we'll make a few minor changes to the package's interface, and then get this released!

PR #137, opened 2021-12-17
Add directory_id and idp_id to directory user, as well as the inactive state constant

logangingerich:
Update from user feedback in this issue: https://github.com/workos-inc/workos-go/issues/136


sheldonvaughn (approved):
Compared to previous PRs, looks good!

PR #139, opened 2021-12-21
Add support for organization parameter

AshGodfrey:
This PR adds support for the organization parameter to initiate SSO.

As part of this, the domain parameter has also been deprecated in favor of organization.

Resolves SDK-355.

### DIFF
--- a/pkg/sso/client.go
+++ b/pkg/sso/client.go
@@ -112,7 +112,7 @@ type GetAuthorizationURLOptions struct {
 	// The unique identifier for a WorkOS Connection.
 	Connection string
 
-	//  The organization Id can be used to initiate SSO.
+	//  The unique identifier for a WorkOS Organization.
 	Organization string
 
 	// The callback URL where your app redirects the user-agent after an
--- a/pkg/sso/client.go
+++ b/pkg/sso/client.go
@@ -112,7 +112,7 @@ type GetAuthorizationURLOptions struct {
 	// The unique identifier for a WorkOS Connection.
 	Connection string
 
-	//  The unique identifier for a WorkOS Organization.
+	// The unique identifier for a WorkOS Organization.
 	Organization string
 
 	// The callback URL where your app redirects the user-agent after an
--- a/pkg/sso/client.go
+++ b/pkg/sso/client.go
@@ -93,7 +93,7 @@ func (c *Client) init() {
 // GetAuthorizationURLOptions contains the options to pass in order to generate
 // an authorization url.
 type GetAuthorizationURLOptions struct {
-	// DEPRECATED please use oranization parameter instead
+	// Deprecated: Please use `Organization` parameter instead.
 	// The app/company domain without without protocol (eg. example.com).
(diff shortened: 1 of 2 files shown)


pkg/sso/client.go
```diff
@@ -111,6 +112,9 @@ type GetAuthorizationURLOptions struct {
 	// The unique identifier for a WorkOS Connection.
 	Connection string
 
+	//  The organization Id can be used to initiate SSO.
```
maxdeviant, line 115: Could we make this comment analogous to the one for `Connection`?

```suggestion
	//  The unique identifier for a WorkOS Organization.
```

pkg/sso/client_test.go
```diff
@@ -64,6 +64,16 @@ func TestClientAuthorizeURL(t *testing.T) {
...
 		},
+		{
+			scenario: "generate url with organization",
+			options: GetAuthorizationURLOptions{
+				Organization: "organization_123",
+				RedirectURI:  "https://example.com/sso/workos/callback",
+				State:        "custom state",
+			},
+			//CHECK THIS EXPECTED URL
+			expected: "https://api.workos.com/sso/authorize?client_id=client_123&organization=organization_123&redirect_uri=https%3A%2F%2Fexample.com%2Fsso%2Fworkos%2Fcallback&response_type=code&state=custom+state",
```
maxdeviant, line 75: This URL looks good to me.

I'm guessing we can remove the comment above?

pkg/sso/client.go
```diff
@@ -93,6 +93,7 @@ func (c *Client) init() {
 // GetAuthorizationURLOptions contains the options to pass in order to generate
 // an authorization url.
 type GetAuthorizationURLOptions struct {
+	// DEPRECATED please use oranization parameter instead
```
maxdeviant, line 96:
```suggestion
	// DEPRECATED Please use `Organization` parameter instead.
```

maxdeviant: It looks like we should be using the `Deprecated:` keyword, according to here:

> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary, but must be kept for compatibility with existing programs. To signal that an identifier should not be used, add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.
>
> &mdash; [source](https://go.dev/blog/godoc#:~:text=Sometimes%20a%20struct,about%20the%20deprecation.)

pkg/sso/client.go
```diff
@@ -111,6 +112,9 @@ type GetAuthorizationURLOptions struct {
 	// The unique identifier for a WorkOS Connection.
 	Connection string
 
+	//  The unique identifier for a WorkOS Organization.
```
maxdeviant, line 115: Looks like there's an extra leading space here:

```suggestion
	// The unique identifier for a WorkOS Organization.
```

maxdeviant (approved):
One last comment, but otherwise I think this is good to merge 👍🏻

PR #142, opened 2022-02-28
Add organization filter to list directories method

AshGodfrey:
Add Organization parameter to List Organizations

### DIFF
--- a/pkg/directorysync/client.go
+++ b/pkg/directorysync/client.go
@@ -501,7 +501,7 @@ func (c *Client) ListDirectories(
 	q.Add("search", opts.Search)
 	q.Add("before", opts.Before)
 	q.Add("after", opts.After)
-	q.Add("organization", opts.OrganizationID)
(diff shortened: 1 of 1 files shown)


pkg/directorysync/client.go
```diff
@@ -447,6 +447,9 @@ type ListDirectoriesOpts struct {
 	// Searchable text for a Directory. Can be empty.
 	Search string
 
+	// Organization ID of a Directory. Can be empty.
+	OrganizationID string
```
mthadley, line 451: We need to also read this value from the options and add it to the URL's query, similar to what we do with the other parameters [here](https://github.com/workos-inc/workos-go/blob/0ba82c57e07e0cab2a6e2dffdb9d99a95a1a798e/pkg/directorysync/client.go#L497).

AshGodfrey: Done, I believe.

mthadley (approved):

PR #144, opened 2022-03-11
Add MFA to SDK

AshGodfrey:


pkg/mfa/client.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	if err = workos.TryGetHTTPError(resp); err != nil {
+		return VerifyResponse{}, err
+	}
+
+	var body VerifyResponse
+	dec := json.NewDecoder(resp.Body)
+	err = dec.Decode(&body)
+	return body, err
```
robframpton, line 272: Wondering what the output would be if one of these errors are returned? https://workos-test.com/docs/mfa/guide/verify. Specifically the `Already Verified` and `Expired` errors.

pliao28: It looks like it might return null id and valid attribute based on VerifyResponse struct. When the call returns an error (such as 422 response when it's already verified) the json reponse body will have a 'code' and 'message' field for the error code and error message respectively. You can nest the VerifyResponse between VerifyResponseSuccess and VerifyResponseError, one will return on Error and other will return on Success 201

AshGodfrey: Right now when false it returns `challenge` and `valid` ----

 IE: 

```
{
  "challenge": {
    "object": "authentication_challenge",
    "id": "auth_challenge_01FVYZWQTZQ5VB6BC5MPG2EYC5",
    "created_at": "2022-02-15T15:26:53.274Z",
    "updated_at": "2022-02-15T15:26:53.274Z",
    "expires_at": "2022-02-15T15:36:53.279Z",
    "authentication_factor_id": "auth_factor_01FVYZ5QM8N98T9ME5BCB2BBMJ"
  },
  "valid": false
}
```

obviously not the desired behavior so I am digging into how exactly to solve this in Go.

AshGodfrey: Hmm, I'm not sure exactly what I'm doing wrong but I'm always getting a 201 created when I make this request to verify (even on postman) which means in reply I'm getting what I return: 

```{
    "challenge": {
        "object": "authentication_challenge",
        "id": "idhere",
        "created_at": "2022-03-23T22:23:46.106Z",
        "updated_at": "2022-03-23T22:23:46.106Z",
        "authentication_factor_id": "factorhere"
    },
    "valid": false
}```

I will put some time on your calendar tomorrow @pliao28 to figure out what I'm missing to get this solved

AshGodfrey: Right now if there's an error it returns the literal error @pliao28 & @robframpton : 
IE - 

```validation failed: 422 Unprocessable Entity: request id "---": The authentication challenge '---' has already been verified.```

I will work on changing to show the error in JSON format.

robframpton: Perfect, thanks @AshGodfrey!

I also fixed the formatting issue in the `master` branch so if you rebase the `gofmt` should work now.

AshGodfrey:
Ugh @robframpton I ran `gofmt` and am still hitting this. Do you have any thoughts?

robframpton:
@AshGodfrey hmm strange, let me pull down the branch locally and see what happens 🤔

robframpton:
@AshGodfrey alrighty I think we're good now 👍🏻

robframpton (approved):

PR #147, opened 2022-03-23
Update Go to 1.17

robframpton:
Trying this separately from https://github.com/workos/workos-go/pull/144

Updates Go version to 1.17.5 in Semaphore and runs `gofmt`.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -12,9 +12,7 @@ blocks:
         - name: gofmt
           commands:
             - checkout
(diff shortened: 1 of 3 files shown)


mthadley (approved):
LGTM!

Sort of related thought: I wonder if we'll want to setup up a build-matrix to make sure we are still backwards compatible with older Go versions.

robframpton:
@mthadley yeah we should probably run unit tests for older versions as well. Not sure how far back we need to go.

Issue #151, opened 2022-04-18
Utility function for getting error code


calufornia:
I notice that there's a [function](https://github.com/workos/workos-go/blob/main/pkg/workos_errors/errors.go) for determining if an error is a 400, but I've encountered errors where the HTTP status code is 422, for example when an email is malformed. It would be great if there were a utility method that could return the status code of a workos `HTTPError` - right now I'm just looking at the prefix of the error message and seeing if it start with 422.

An alternative is to expose the [HTTPError struct](https://github.com/workos/workos-go/blob/main/pkg/workos_errors/errors.go) so that the client can perform a type assertion.

brunocalogero:
Hey @calufornia, this has been addressed and merged very recently by one of my team-mates :) https://github.com/workos/workos-go/pull/152

The HTTPError struct is exposed

mthadley:
As @brunocalogero pointed out, the `HTTPError` struct is now public, so this should hopefully allow you to introspect on the particular error code.

I'l close this for now, but if you hit any more issues, feel free to open a new issue. Thanks!

PR #152, opened 2022-05-06
Expose HTTPError struct

AgustinBanchio:
# Context

The HTTPError struct is part of the API response and as such should be exposed.
If it's hidden in an internal package, then error handling for WorkOS API users becomes a problem.

The only way to get the error code is to parse the error string, and that implementation is not ideal and could become invalid whenever the Error method changes.

Currently, there's only a public bad request checker in the workos_errors pkg. Many times, invalid requests have a different code than just 400 (we receive many 422, for example).

# Description

This PR moves the `HTTPError` struct definition, methods, helping functions and tests from the internal `workos` pkg to the public `workos_errors` pkg.

It also updates all their usages as well.

This PR would also close the following issue: https://github.com/workos/workos-go/issues/151

### DIFF
--- a/pkg/audittrail/client.go
+++ b/pkg/audittrail/client.go
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/workos/workos-go/pkg/workos_errors"
+
 	"github.com/google/go-querystring/query"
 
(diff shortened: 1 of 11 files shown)


ams11 (approved):
👍 Looks great - thanks for addressing this, @AgustinBanchio!

PR #158, opened 2022-06-08
Map all available fields in directory group

AshGodfrey:


pkg/directorysync/client.go
```diff
@@ -214,6 +214,21 @@ type Group struct {
 
 	// The Group's name.
 	Name string `json:"name"`
+
+	// The Groups's unique identifier assigned by the Directory Provider.
```
maxdeviant, line 218:
```suggestion
	// The Group's unique identifier assigned by the Directory Provider.
```

sheldonvaughn (approved):

PR #160, opened 2022-07-06
Update endpoints used for MFA operations

AshGodfrey:
This PR updates the endpoints used for two of the MFA operations:

- `ChallengeFactor`
- `VerifyChallenge`

The VerifyFactor method is now deprecated and has been replaced with VerifyChallenge. VerifyFactor will be removed in the next major version.

Resolves SDK-546.

I followed Godoc's deprecation:
https://go.dev/blog/godoc

### DIFF
--- a/pkg/mfa/client.go
+++ b/pkg/mfa/client.go
@@ -245,6 +245,14 @@ func (c *Client) ChallengeFactor(
 
 }
 
+// Deprecated: Use VerifyChallenge instead.
+func (c *Client) VerifyFactor(
+	ctx context.Context,
+	opts VerifyOpts,
+) (interface{}, error) {
+	return VerifyChallenge(ctx, opts)
+}
+
 // Verifies the one time password provided by the end-user.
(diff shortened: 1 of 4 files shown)


pkg/mfa/client.go
```diff
@@ -263,7 +270,7 @@ func (c *Client) VerifyFactor(
 	})
 	responseBody := bytes.NewBuffer(postBody)
 
-	endpoint := fmt.Sprintf("%s/auth/factors/verify", c.Endpoint)
+	endpoint := fmt.Sprintf("%s/auth/challenges/%s/verify", c.Endpoint, opts.AuthenticationChallengeID)
```
maxdeviant, line 273: Could we also remove the `authentication_challenge_id` field from the `postBody` above?

AshGodfrey: Done, sorry for the delay

maxdeviant (approved):

PR #162, opened 2022-07-14
Bump version to v1.13.0

willmanduffy:
* #161
* #160


maxdeviant:
It looks like this PR should go in the release notes as well: https://github.com/workos/workos-go/pull/160

PR #164, opened 2022-08-22
Add Audit Logs Client

jonatascastro12:
Add Audit Logs Client with the following functions:
- CreateEvent
- CreateExport
- GetExport

### DIFF
--- a/pkg/auditlogs/client.go
+++ b/pkg/auditlogs/client.go
@@ -35,11 +35,12 @@ type Client struct {
 	// to http.Client.
 	HTTPClient *http.Client
 
-	// The endpoint used to request Workos. Defaults to
-	// https://api.workos.com/audit_logs/events.
+	// The endpoint used to request WorkOS AuditLog events creation endpoint.
+	// Defaults to https://api.workos.com/audit_logs/events.
 	EventsEndpoint string
 
-	// https://api.workos.com/audit_logs/exports.
+	// The endpoint used to request WorkOS AuditLog events creation endpoint.
+	// Defaults to https://api.workos.com/audit_logs/exports.
 	ExportsEndpoint string
 
 	// The function used to encode in JSON. Defaults to json.Marshal.
@@ -55,6 +56,7 @@ type AuditLogEventOpts struct {
 
 	// Event payload
 	Event Event `json:"event" binding:"required"`
+
 	// If no key is provided or the key is empty, the key will not be attached
 	// to the request.
 	IdempotencyKey string `json:"-"`
@@ -68,7 +70,6 @@ type Event struct {
 	Version int `json:"version,omitempty"`
 
 	// The time when the audit trail occurred.
-	//
 	// Defaults to time.Now().
 	OccurredAt time.Time `json:"occurred_at"`
 
--- a/pkg/auditlogs/client.go
+++ b/pkg/auditlogs/client.go
@@ -27,7 +27,7 @@ const (
 
 // Client represents a client that performs auditlogs requests to WorkOS API.
 type Client struct {
-	// The WorkOS api key. It can be found in
+	// The WorkOS API key. It can be found in
 	// https://dashboard.workos.com/api-keys.
 	APIKey string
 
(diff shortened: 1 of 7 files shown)


pkg/auditlogs/client_test.go
```diff
@@ -0,0 +1,316 @@
...
+		_, err := GetExport(context.TODO(), GetExportOpts{})
+		require.Error(t, err)
+	})
+}
+
+type defaultTestHandler struct {
+	header *http.Header
+}
+
+// Unit test to hit the prod api directly. Uncomment and provide an APIKey to
```
AshGodfrey, line 283: Is this unit test necessary? I don't think any of our other tests have this, so I'm just wondering if something is unique here that requires this or if it was a specific ask (or if other unit tests should be improved)

pkg/auditlogs/client.go
```diff
@@ -0,0 +1,280 @@
...
+
+// Constants that enumerate the available orders.
+const (
+	Asc  Order = "asc"
+	Desc Order = "desc"
+)
+
+// Client represents a client that performs auditlogs requests to WorkOS API.
+type Client struct {
+	// The WorkOS api key. It can be found in
```
AshGodfrey, line 30: nitpick but our similar comments are inline with links (I think there's an argument for putting the links in their own section like the go docs suggest, but right now this is all just inline)

pkg/auditlogs/client.go
```diff
@@ -0,0 +1,280 @@
...
+		return CreateExportResponse{}, err
+	}
+
+	var body CreateExportResponse
+	dec := json.NewDecoder(res.Body)
+	err = dec.Decode(&body)
+	return body, err
+}
+
+// GetExport creates an export of Audit Log events. You can specify some filters.
```
AshGodfrey, line 246: Is there a spot where I'd be able to see what is filtered? (Or should this comment be eventually updated with a not yet published doc?)

pkg/auditlogs/client_test.go
jonatascastro12: So I used this code as a reference:
https://github.com/workos/workos-go/blob/dfb61a3e3268f2e8843cfe87d30e4431b1685108/pkg/audittrail/client_test.go#L249-L269

It was actually useful when I tested live. I found a few errors and I was able to fix it.

Should we remove it?

pkg/auditlogs/client.go
jonatascastro12: Oh... That's a missing copy/paste. I'll update it, but it will eventually be updated after we publish the docs. 😄

AshGodfrey:
Just waiting on any information regarding the unit test, but otherwise looks good @jonatascastro12

pkg/auditlogs/client.go
jonatascastro12: So we discussed it on Slack and I'm following the convention from other parts of the repo where there's a limit of approx. 70-80 chars for breaking the line in field comments.

jonatascastro12:
@AshGodfrey I will remove that commented test, but in another moment I should open a discussion about end-to-end tests in the SDKs.

AshGodfrey (approved):

pkg/auditlogs/client.go
```diff
@@ -0,0 +1,281 @@
...
+
+// Constants that enumerate the available orders.
+const (
+	Asc  Order = "asc"
+	Desc Order = "desc"
+)
+
+// Client represents a client that performs auditlogs requests to WorkOS API.
+type Client struct {
+	// The WorkOS api key. It can be found in
```
maxdeviant, line 30:
```suggestion
	// The WorkOS API key. It can be found in
```

pkg/auditlogs/client.go
```diff
@@ -0,0 +1,281 @@
...
+	// https://dashboard.workos.com/api-keys.
+	APIKey string
+
+	// The http.Client that is used to post audit trail events to WorkOS. Defaults
```
maxdeviant, line 34:
```suggestion
	// The http.Client that is used to post Audit Log events to WorkOS. Defaults
```

PR #167, opened 2022-09-06
Bump Version to 1.15.0

AshGodfrey:
Adds idempotency key support


maxdeviant:
Could we change the PR title to include the version number? This makes it easier to see version bumps in Git history after the PR has been merged.

We also don't need to indicate the changes in this version in the title (e.g., no need to mention that this version includes idempotency key support), as this can be in the PR description and the release notes.

jonatascastro12 (approved):

PR #169, opened 2022-09-09
Add PrimaryEmail method for Directory User

AshGodfrey:
Adds a method to get the primary email of a directory user.

### DIFF
--- a/pkg/directorysync/directorysync.go
+++ b/pkg/directorysync/directorysync.go
@@ -74,7 +74,7 @@ func DeleteDirectory(
 	return DefaultClient.DeleteDirectory(ctx, opts)
 }
 
-// Primary Email is a method for finding a user's primary email (when applicable)
+// PrimaryEmail is a method for finding a user's primary email (when applicable)
 func (r User) PrimaryEmail() (string, error) {
 	for _, v := range r.Emails {
 		if v.Primary {
--- a/pkg/directorysync/directorysync.go
+++ b/pkg/directorysync/directorysync.go
@@ -3,6 +3,7 @@ package directorysync
 
 import (
 	"context"
+	"errors"
 )
 
 // DefaultClient is the client used by SetAPIKey and Directory Sync functions.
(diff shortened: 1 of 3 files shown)


pkg/directorysync/directorysync_test.go
```diff
@@ -296,3 +296,128 @@ func getDirectoryTestHandler(w http.ResponseWriter, r *http.Request) {
...
+					},
+				},
+				State:            Active,
+				RawAttributes:    json.RawMessage(`{"foo":"bar"}`),
+				CustomAttributes: json.RawMessage(`{"foo":"bar"}`),
+			},
+			expected: "firstprimaryemail@foo-corp.com",
+		},
+		{
+			scenario: "No primary emails returns null",
```
blairlunceford, line 365: It looks like we're returning an error rather than `null` here. Can we rename these tests to indicate an error is returned as well as an empty string?

blairlunceford (changes requested):

pkg/directorysync/directorysync.go
```diff
@@ -72,3 +73,13 @@ func DeleteDirectory(
 ) error {
 	return DefaultClient.DeleteDirectory(ctx, opts)
 }
+
+// PrimaryEmail is a method for finding a user's primary email (when applicable)
+func (r User) PrimaryEmail() (string, error) {
```
blairlunceford, line 78: Should this logic be in the client.go file, like the other methods? And then called via `DefaultClient`?

AshGodfrey: I could be wrong here, but my gut said no. Client performs Directory Sync requests to the WorkOS API, but primary email doesn't actually make a request to the API, it just takes information that already exists from the User type and distills it.

maxdeviant: If I understand this correctly, I do think that the `PrimaryEmail` should be a method on the `User` object rather than being tied with the methods on the client that actually call the API.

blairlunceford: Ahh ok, that makes sense. It's just we only have methods that use the client that connects to the API, so that's why we have that logic in the client now.

blairlunceford (approved):

PR #172, opened 2022-10-10
Add support for DeleteFactor and GetFactor methods

AshGodfrey:
Add support for delete factor and get factor methods
Add tests for delete factor and get factor methods

### DIFF
--- a/pkg/mfa/client.go
+++ b/pkg/mfa/client.go
@@ -160,7 +160,7 @@ type DeleteFactorOpts struct {
 
 type GetFactorOpts struct {
 	// ID of the factor.
-	ID string
+	AuthenticationFactorID string
 }
 
 // Create an Authentication Factor.
@@ -305,7 +305,7 @@ func (c *Client) VerifyChallenge(
 
 }
 
-// DeleteFactor deletes a Factor.
+// Deletes an authentication factor.
 func (c *Client) DeleteFactor(
 	ctx context.Context,
 	opts DeleteFactorOpts,
@@ -340,7 +340,7 @@ func (c *Client) DeleteFactor(
 	return workos_errors.TryGetHTTPError(res)
 }
 
-// GetFactor gets a Factor.
+// Retrieves an authentication factor.
 func (c *Client) GetFactor(
 	ctx context.Context,
 	opts GetFactorOpts,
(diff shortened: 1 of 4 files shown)


pkg/mfa/client.go
```diff
@@ -153,6 +153,16 @@ type RawVerifyResponse struct {
 	VerifyResponseError
 }
 
+type DeleteFactorOpts struct {
+	// ID of factor to be deleted
+	ID string
```
blairlunceford, line 158: It looks like the convention is to have a more descriptive name - could we use `AuthenticationFactorID` instead for both of these new methods?

pkg/mfa/client.go
```diff
@@ -294,3 +304,80 @@ func (c *Client) VerifyChallenge(
 	}
 
 }
+
+// DeleteFactor deletes a Factor.
```
blairlunceford, line 308: nit: change comment to `Deletes an authentication factor` to be in line with other methods.

pkg/mfa/client.go
```diff
@@ -294,3 +304,80 @@ func (c *Client) VerifyChallenge(
...
+	res, err := c.HTTPClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	return workos_errors.TryGetHTTPError(res)
+}
+
+// GetFactor gets a Factor.
```
blairlunceford, line 343: nit: change comment to `Retrieves an authentication factor` to be in line with other methods.

blairlunceford (changes requested):

AshGodfrey:
Going to need to check why this failing.

AshGodfrey:
Also I switched the ID name as requested, but a lot of our other go functions do just use ID. Not sure if we want to keep it inline with other functions or go with the more descriptive name.

blairlunceford:
> Also I switched the ID name as requested, but a lot of our other go functions do just use ID. Not sure if we want to keep it inline with other functions or go with the more descriptive name.

Thanks! Yes, but this pattern will follow the `ChallengeOpts` and `VerifyOpts` structs in this file, so I think it's good to just make the change here.

blairlunceford (approved):

PR #175, opened 2022-10-12
Add `CreatedAt` and `UpdatedAt` to `User`

dmytro-cyberhaven:


internal/workos/workos.go
```diff
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v1.16.0"
+	Version = "v1.17.0"
```
maxdeviant, line 5: Could we omit the version change from the PR?

We bump the version separately when we're ready to cut a new SDK release.

dmytro-cyberhaven: Done

blairlunceford (approved):

PR #180, opened 2022-11-09
sso: Add GetLoginHandler to sso.Client

drewfradette:
The `sso.Login` function was only available for `sso.DefaultClient` meaning the only way to use it was to set `sso.DefaultClient` first.

By moving the handler creation into `sso.Client`, any `Client` can now create a login handler that respects the client instances values.


jonatascastro12 (approved):
lgtm

PR #181, opened 2022-11-10
Clean inconsistencies on Audit Log Client

jonatascastro12:
Clean some inconsistencies on the Audit Log Client:

- Every other piece of the Go SDK keeps correct Go casing with abbreviations (e.g. ID, URL, ConnectionID). Audit Logs stuff breaks this (OrganizationId, Id, Url)
- Every other piece of the Go SDK follows patterns to how objects and types are named. E.g. GetOrganizations() accepts GetOrganizationsOpts, and returns Organization, and so does every other function. However, in Audit Logs there is no pattern. E.g. CreateEvent() accepts AuditLogEventOpts.
- Every other piece of the Go SDK specifies enum-like string types. For Audit Logs enums are just marked as string. The IDE gives you no intellisense on these.
- Audit Logs Go SDK misses a common interface for an audit logs export. There is a GetExport() function, but it returns GetExportResponse.  There is also CreateExport() function, which returns CreateExportResponse, which is exactly the same object as the previous function returns. There is no common Export or AuditLogsExport interface like in other SDKs. - The type signatures make it look like they return different things, when in reality they are not. It’s a bit confusing to navigate the SDK even in the IDE. It also makes it harder to document things in the docs; it’s very unhuman-like to say that something gives you a GetExportResponse.

Resolves LOG-540

### DIFF
--- a/pkg/auditlogs/README.md
+++ b/pkg/auditlogs/README.md
@@ -21,12 +21,12 @@ func main() {
 	auditlogs.SetAPIKey("my_api_key")
 
 	// Wherever you need to publish an audit log event:
-	err := auditlogs.CreateEvent(ctx.Background(), auditlogs.AuditLogEventOpts{
+	err := auditlogs.CreateEvent(ctx.Background(), auditlogs.CreateEventOpts{
 		Organization: "org_8899300049990088",
 		Event: Event{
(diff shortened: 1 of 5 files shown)


vladmoroz:
Wooohoo

PR #187, opened 2022-11-15
Standardise return type of VerifyChallenge and coalesce errors into Error

aultimus:
Prefer strongly typed return types to interface{} and map[string]interface{}. Don't use two separate Error like return values.

### DIFF
--- a/pkg/mfa/client.go
+++ b/pkg/mfa/client.go
@@ -134,7 +134,7 @@ type VerifyOpts struct {
 
 type VerifyResponse struct {
 	// Return details of the request
-	Challenge ChallengeResponse
+	Challenge ChallengeResponse `json:"challenge"`
 
 	// Boolean returning if request is valid
 	Valid bool `json:"valid"`
--- a/pkg/mfa/client.go
+++ b/pkg/mfa/client.go
@@ -288,7 +288,7 @@ type VerificationResponseError struct {
 	Message string
 }
 
-func (r *VerificationResponseError) Error() string {
+func (r VerificationResponseError) Error() string {
 	return fmt.Sprintf("code: %s. message: %v", r.Code, r.Message)
 }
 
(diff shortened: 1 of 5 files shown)


jonatascastro12 (approved):
lgtm

aultimus:
Need to release this as a major version update in semver terms as this is a breaking change. I see a test Makefile target but no instructions on usage. @maxdeviant do i run `make release` within this pr or merge this then run `make release afterwards  on main?

maxdeviant:
> Need to release this as a major version update in semver terms as this is a breaking change. I see a test Makefile target but no instructions on usage. @maxdeviant do i run `make release` within this pr or merge this then run `make release afterwards on main?

The version bump should be done as a separate PR.

I'll DM you the internal docs we have on the SDK release process.

pkg/mfa/client.go
```diff
@@ -263,11 +263,20 @@ func (c *Client) VerifyFactor(
 	return VerifyChallenge(ctx, opts)
 }
 
+type VerificationResponseError struct {
+	Code    string
+	Message string
+}
+
+func (r *VerificationResponseError) Error() string {
+	return fmt.Sprintf("code: %s. message: %v", r.Code, r.Message)
```
drewfradette, line 272: Adding more context to the message would be nice probably?
```suggestion
	return fmt.Sprintf("mfa verification failed: %s (code: %s)", err.Message, err.Code)
```

AshGodfrey:
I can go ahead and make these changes @drewfradette - I'm going to make the changes requested in #188 and merge into this, and then we can continue review + merge this PR to the 2.0 branch from there, I think?

pkg/mfa/client.go
```diff
@@ -87,24 +97,34 @@ type EnrollResponse struct {
 	UpdatedAt string `json:"updated_at"`
 
 	// The type of request either 'sms' or 'totp'
-	Type string `json:"type"`
+	Type FactorType `json:"type"`
 
-	// Details of the totp response will be 'null' if using sms
-	Totp map[string]interface{} `json:"totp"`
+	// Details of the totp response will be 'null' if using sms\
```
drewfradette, line 102:
```suggestion
	// Details of the totp response will be 'null' if using sms
```

drewfradette (approved):
one stray `\` but 🚀

PR #189, opened 2022-11-16
mfa: Make client functions more consistent.

drewfradette:
Noticed a few additional bits while reviewing #188 that were outside the scope of the PR so I split them out here:

* All `mfa.Client` methods now utilize provided `ctx`.
* Removed `log.Panic` in favour of returning errors.
* All `mfa.Client` methods now utilize the client's `HTTPClient`.
* Added `defer`s for response bodies.
* Utilize `http.Method*` constants instead of strings.
* Have `Client.VerifyFactor` call it's receiver, not `DefaultClient`.

### DIFF
--- a/pkg/mfa/client.go
+++ b/pkg/mfa/client.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"net/http"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


drewfradette:
Retargeted the base to the `2.0.0-staging` branch instead of `main`.

jonatascastro12 (approved):

PR #207, opened 2023-05-23
Refactor Directory Sync and SSO to support order parameter

AshGodfrey:
## Description

Directory Sync and SSO list endpoints couldn't accept the order parameter. This refactor allows these functions to send that and future parameters without needing to manually add them. 

## Documentation

```
[ ] Yes
```


pkg/directorysync/client.go
```diff
@@ -137,22 +137,22 @@ type User struct {
 // ListUsersOpts contains the options to request provisioned Directory Users.
 type ListUsersOpts struct {
 	// Directory unique identifier.
-	Directory string
+	Directory string `url:"directory", omitempty`
 
 	// Directory Group unique identifier.
-	Group string
+	Group string `url:"group", omitempty`
```
drewfradette, line 143: struct tags are weird magic strings but they are pretty picky about the format.

```suggestion
	Group string `url:"group,omitempty"`
```
(this is the case with a decent number of the tags, just commenting on this one though!)

pkg/directorysync/client.go
```diff
@@ -137,22 +137,22 @@ type User struct {
...
+	Order Order `url:"order,omitempty"`
 
-	// Pagination cursor to receive records before a provided Directory ID.
-	Before string
+	// Pagination cursor to receive records before a provided Event ID.
+	Before string `url:"before,omitempty"`
 
-	// Pagination cursor to receive records after a provided Directory ID.
-	After string
+	// Pagination cursor to receive records after a provided Event ID.
```
drewfradette, line 154: is the change of `Directory` to `Event` intended here? (it's in a few places but just gonna comment here)

drewfradette (changes requested):
I've never actually seen the go-querystring package before, solid refactor!

A few small snags but ping me again whenever!

pkg/sso/client.go
```diff
@@ -456,25 +455,25 @@ func (c *Client) GetConnection(
...
-	Limit int
+	Limit int `url:"limit"`
 
 	// The order in which to paginate records.
-	Order Order
+	Order Order `url:"order,omitempty"`
 
-	// Pagination cursor to receive records before a provided Connection ID.
-	Before string
+	// Pagination cursor to receive records before a provided Event ID.
```
drewfradette, line 472: This should be `Connection` still I believe!

pkg/sso/client.go
```diff
@@ -456,25 +455,25 @@ func (c *Client) GetConnection(
...
+	Before string `url:"before,omitempty"`
 
-	// Pagination cursor to receive records after a provided Connection ID.
```
drewfradette, line 476: This should be `Connection` still I believe!

drewfradette (approved):
two comment nits but otherwise lgtm

PR #208, opened 2023-05-31
Add Events API beta

alisherry:
## Description
Adding beta support for the Events API. 

## Documentation

```
[ X ] Yes
```

### DIFF
--- a/pkg/events/client.go
+++ b/pkg/events/client.go
@@ -1,0 +1,135 @@
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/google/go-querystring/query"
+	"github.com/workos/workos-go/v2/pkg/workos_errors"
+
+	"github.com/workos/workos-go/v2/internal/workos"
+	"github.com/workos/workos-go/v2/pkg/common"
+)
+
+// ResponseLimit is the default number of records to limit a response to.
+const ResponseLimit = 10
+
+// Client represents a client that performs Event requests to the WorkOS API.
+type Client struct {
+	// The WorkOS API Key. It can be found in https://dashboard.workos.com/api-keys.
+	APIKey string
+
+	// The http.Client that is used to get Event records from WorkOS.
+	// Defaults to http.Client.
+	HTTPClient *http.Client
+
+	// The endpoint to WorkOS API. Defaults to https://api.workos.com.
+	Endpoint string
+
+	once sync.Once
+}
+
+func (c *Client) init() {
+	if c.HTTPClient == nil {
+		c.HTTPClient = &http.Client{Timeout: 10 * time.Second}
+	}
+
+	if c.Endpoint == "" {
+		c.Endpoint = "https://api.workos.com"
+	}
+}
+
+
+// Event contains data about a particular Event.
+type Event struct {
+	// The Event's unique identifier.
+	ID string `json:"id"`
+
+	// The type of Event.
+	Event string `json:"event"`
+
+	// The Event's data in raw encoded JSON.
+	Data json.RawMessage `json:"data"`
+
+	// The Event's created at date.
+	CreatedAt string `json:"created_at"`
+}
+
+// ListeEventsOpts contains the options to request provisioned Events.
+type ListeEventsOpts struct {
+	// Filter to only return Events of particular types.
+	Events string[] `url:"events,omitempty"`
+
+	// Maximum number of records to return.
+	Limit int `url:"limit"`
+
+	// Pagination cursor to receive records after a provided Event ID.
+	After string `url:"after,omitempty"`
+
+	// Date range start for stream of Events.
+	RangeStart string `url:"after,omitempty"`
+
+	// Date range end for stream of Events.
+	RangeEnd string `url:"after,omitempty"`
+}
+
+// ListEventsResponse describes the response structure when requesting
+// Events.
+type ListEventsResponse struct {
+	// List of Events.
+	Data []Event `json:"data"`
+
+	// Cursor pagination options.
+	ListMetadata common.ListMetadata `json:"listMetadata"`
+}
+
+// ListEvents gets a list of Events.
+func (c *Client) ListEvents(
+	ctx context.Context,
+	opts ListeEventsOpts,
+) (ListeEventsOpts, error) {
+	c.once.Do(c.init)
+
+	endpoint := fmt.Sprintf("%s/events", c.Endpoint)
+	req, err := http.NewRequest(
+		http.MethodGet,
+		endpoint,
+		nil,
+	)
+	if err != nil {
+		return ListEventsResponse{}, err
+	}
+
+	req = req.WithContext(ctx)
+	req.Header.Set("Authorization", "Bearer "+c.APIKey)
+	req.Header.Set("Content-Type", "application/json")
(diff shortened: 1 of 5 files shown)


pkg/events/client.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/google/go-querystring/query"
+	"github.com/workos/workos-go/v2/pkg/workos_errors"
+
+	"github.com/workos/workos-go/v2/internal/workos"
+	"github.com/workos/workos-go/v2/pkg/common"
+)
+
+// ResponseLimit is the default number of records to limit a response to.
+const ResponseLimit = 10
```
tribble, line 19: Is this the default limit we want?

pkg/events/client.go
```diff
@@ -0,0 +1,134 @@
...
+	ID string `json:"id"`
+
+	// The type of Event.
+	Event string `json:"event"`
+
+	// The Event's data in raw encoded JSON.
+	Data json.RawMessage `json:"data"`
+
+	// The Event's created at date.
+	CreatedAt string `json:"created_at"`
```
tribble, line 58: Should this be a string? We seem to use different types in different parts of the SDK. Audit logs [uses `time.Time`](https://github.com/workos/workos-go/blob/ea6316d005c5c9fb4bf0fe03277edf0d9ef32746/pkg/auditlogs/client.go#L74) but organizations [uses a `string`](https://github.com/workos/workos-go/blob/ea6316d005c5c9fb4bf0fe03277edf0d9ef32746/pkg/organizations/client.go#L88).

pkg/events/client.go
```diff
@@ -0,0 +1,134 @@
...
+type GetEventsResponse struct {
+	// List of Events.
+	Data []Event `json:"data"`
+
+	// Cursor pagination options.
+	ListMetadata common.ListMetadata `json:"listMetadata"`
+}
+
+// GetEvents gets a list of Events.
+func (c *Client) GetEvents(
```
tribble, line 90: We seem to use `Get*` for single object loads. Should this be `ListEvents`?

pkg/events/client_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+			events, err := client.GetEvents(context.Background(), test.options)
+			if test.err {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, test.expected, events)
+		})
+	}
```
tribble, line 66: I don't quite understand why we need to go over this range? It seems unnecessary?

pkg/events/client_test.go
```diff
@@ -0,0 +1,104 @@
...
+func TestGetEvents(t *testing.T) {
+	tests := []struct {
+		scenario string
+		client   *Client
+		options  GetEventsOpts
+		expected GetEventsResponse
+		err      bool
+	}{
+		{
+			scenario: "Request without API Key returns an error",
```
tribble, line 24: Is it worth adding a happy path test?

pkg/events/events_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/workos/workos-go/v2/pkg/common"
+)
+
+func TestEventsGetEvents(t *testing.T) {
```
tribble, line 14: Ah, this actually seems like a sufficient happy path test.

tribble (commented):
I think we actually want this to be `ListEvents` to be in line with the naming of our other SDK methods.

pkg/events/client.go
alisherry: Yeah considered using 100 here, but it seems to be the other default response limit across our entire SDK so want to keep it consistent.

pkg/events/client.go
alisherry: Hmm, yeah we could use `time.Time`, thought we were using `string` across the board but Time fits better here.

pkg/events/client.go
```diff
@@ -0,0 +1,134 @@
...
+	Events []string `url:"events,omitempty"`
+
+	// Maximum number of records to return.
+	Limit int `url:"limit"`
+
+	// Pagination cursor to receive records after a provided Event ID.
+	After string `url:"after,omitempty"`
+
+	// Date range start for stream of Events.
+	RangeStart string `url:"after,omitempty"`
```
ospahiu, line 73:
```suggestion
	RangeStart string `url:"rangeStart,omitempty"`
```

pkg/events/client.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Cursor pagination options.
+	ListMetadata common.ListMetadata `json:"listMetadata"`
+}
+
+// GetEvents gets a list of Events.
+func (c *Client) GetEvents(
+	ctx context.Context,
+	opts GetEventsOpts,
+) (GetEventsResponse, error) {
```
ospahiu, line 93: I wonder if we should following a pointer to a return value model:
`) (*GetEventsResponse, error) {`

and returning `nil` in the error case, what does the rest of the SDK do?:

```
if err != nil {
	return nil, err
}	
```

pkg/events/client.go
```diff
@@ -0,0 +1,134 @@
...
+
+	req = req.WithContext(ctx)
+	req.Header.Set("Authorization", "Bearer "+c.APIKey)
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", "workos-go/"+workos.Version)
+	if opts.Limit == 0 {
+		opts.Limit = ResponseLimit
+	}
+
+	v, err := query.Values(opts)
```
ospahiu, line 114:
```suggestion
	queryValues, err := query.Values(opts)
```
nit: more descriptive?

pkg/events/client_test.go
ospahiu: +1

pkg/events/client_test.go
```diff
@@ -0,0 +1,104 @@
+package events
```
ospahiu, line 1: Can we add a range start and end test?

pkg/events/client.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// The Event's data in raw encoded JSON.
+	Data json.RawMessage `json:"data"`
+
+	// The Event's created at date.
+	CreatedAt string `json:"created_at"`
+}
+
+// GetEventsOpts contains the options to request provisioned Events.
+type GetEventsOpts struct {
```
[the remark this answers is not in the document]
alisherry: I think its usually preferred to use non pointer types for primitives. Double checked and it seems we omit the use of pointers across the SDK

pkg/events/client.go
alisherry: It seems we actually return an empty response struct instead of nil across the SDK, probably makes sense to continue the pattern (although I'm open to following the pattern ^^ but  we should make it uniform across the SDK

AshGodfrey: FWIW, if it's a good practice - we counld add this to the next major release as an update. I was going to start on some V3 changes on a new branch.

pkg/events/client.go
AshGodfrey: I believe this limit in general is inconsistent across SDKs (it might be 20 in python), so it's worth examining if they all need to be updated.

PR #210, opened 2023-06-23
Use snake case for Events range params

amygdalama:
## Description
Use snake case for range params in the Events API. This makes Events consistent with the rest of our API.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/events/client_test.go
+++ b/pkg/events/client_test.go
@@ -42,7 +42,7 @@ func TestListEvents(t *testing.T) {
 		require.Equal(t, expectedResponse, events)
 	})
 
-	t.Run("ListEvents succeeds to fetch Events with a RangeStart and RangeEnd", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/events/client_test.go
```diff
@@ -42,7 +42,7 @@ func TestListEvents(t *testing.T) {
 		require.Equal(t, expectedResponse, events)
 	})
 
-	t.Run("ListEvents succeeds to fetch Events with a startRange and endRange", func(t *testing.T) {
+	t.Run("ListEvents succeeds to fetch Events with a RangeStart and RangeEnd", func(t *testing.T) {
```
alisherry, line 45: `range_start` and `range_end`?

amygdalama: ah yeah I went back and forth on that! ready for another look.

alisherry (approved):

PR #218, opened 2023-08-07
User management: Get User

jonatascastro12:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -1,0 +1,65 @@
+package users
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+)
+
+// ResponseLimit is the default number of records to limit a response to.
+const ResponseLimit = 10
+
+// Order represents the order of records.
+type Order string
+
+// Constants that enumerate the available orders.
+const (
+	Asc  Order = "asc"
+	Desc Order = "desc"
+)
+
+// Client represents a client that fetch SSO data from WorkOS API.
+type Client struct {
+	// The WorkOS api key. It can be found in
+	// https://dashboard.workos.com/api-keys.
+	//
+	// REQUIRED.
+	APIKey string
+
+	// The WorkOS Client ID (eg. client_01JG3BCPTRTSTTWQR4VSHXGWCQ).
+	//
+	// REQUIRED.
+	ClientID string
+
+	// The endpoint to WorkOS API.
+	//
+	// Defaults to https://api.workos.com.
+	Endpoint string
+
+	// The http.Client that is used to send request to WorkOS.
+	//
+	// Defaults to http.Client.
+	HTTPClient *http.Client
+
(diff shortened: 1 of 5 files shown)


pkg/users/client.go
```diff
@@ -0,0 +1,183 @@
...
+	// Defaults to http.Client.
+	HTTPClient *http.Client
+
+	// The function used to encode in JSON. Defaults to json.Marshal.
+	JSONEncode func(v interface{}) ([]byte, error)
+
+	once sync.Once
+}
+
+func (c *Client) init() {
```
AshGodfrey, line 56: We should probably switch from this method to something more like the new SSO method. Instead of being set in client.go it can be set in users.go. 

A NewClient function is a more idiomatic approach. I'm not sure this is a case where lazy initialization has advantages enough to make the init() make more sense?






```
// NewClient returns a new instance of the Client struct with default values.
func NewClient(apiKey, clientID string) *Client {
	return &Client{
		APIKey:     apiKey,
		ClientID:   clientID,
		Endpoint:   "https://api.workos.com",
		HTTPClient: &http.Client{Timeout: time.Second * 10},
		JSONEncode: json.Marshal,
	}
}
```

jonatascastro12:
Current dependencies on/for this PR:
* users
  * **PR #218** <a href="https://app.graphite.dev/github/pr/workos/workos-go/218" target="_blank">Graphite</a>  👈
    * **PR #219** <a href="https://app.graphite.dev/github/pr/workos/workos-go/219" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/workos/workos-go/218?utm_source=stack-comment).

pkg/users/client.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// The http.Client that is used to send request to WorkOS.
+	//
+	// Defaults to http.Client.
+	HTTPClient *http.Client
+
+	// The function used to encode in JSON. Defaults to json.Marshal.
+	JSONEncode func(v interface{}) ([]byte, error)
+
+	once sync.Once
```
AshGodfrey, line 47: We should be able to pull out once sync.Once.

pkg/users/users.go
```diff
@@ -0,0 +1,24 @@
...
+var (
+	// DefaultClient is the client used by GetAuthorizationURL, GetProfileAndToken and
+	// Login functions.
+	DefaultClient = NewClient("")
+)
+
+// Configure configures the default client that is used by the User management methods
+// It must be called before using those functions.
+func Configure(apiKey string) {
+	DefaultClient.APIKey = apiKey
```
AshGodfrey, line 15: Is this setting the client as expected when called?

jonatascastro12: This should be SetAPIKey. I'll add tests to `users_test.go`

AshGodfrey (approved):

PR #220, opened 2023-08-07
Authenticate with password

AshGodfrey:
## Description

Adds the authenticate with password method

## Documentation

```
[ ] Yes
```


jonatascastro12:
Need to rebase onto `users`

pkg/users/client.go
```diff
@@ -142,3 +144,62 @@ func (c *Client) GetUser(ctx context.Context, opts GetUserOpts) (User, error) {
...
+	Password     string `json:"password"`
+	IPAddress    string `json:"ip_address,omitempty"`
+	UserAgent    string `json:"user_agent,omitempty"`
+	StartSession bool   `json:"start_session,omitempty"`
+	ExpiresIn    int    `json:"expires_in,omitempty"`
+}
+type AuthenticationResponse struct {
+	Session Session `json:"session"`
+	User    User    `json:"user"`
+}
```
jonatascastro12, line 166: Should we add the comments? I'm fine with adding them later too.
I was mostly copying from the existing API reference.

pkg/users/client.go
```diff
@@ -142,3 +144,62 @@ func (c *Client) GetUser(ctx context.Context, opts GetUserOpts) (User, error) {
 
 	return body, err
 }
+
+type Session struct {
+	ID        string `json:"id"`
+	Token     string `json:"token"`
+	CreatedAt string `json:"created_at"`
+	UpdatedAt string `json:"Updated_at"`
```
jonatascastro12, line 152:
```suggestion
	UpdatedAt string `json:"Updated_at"`
```
It should be `ExpiresAt`

pkg/users/client.go
AshGodfrey: I think it's preferred, but if we're being honest in the interest of time I was skipping them for getting functionality in

PR #223, opened 2023-08-07
Add RemoveUserFromOrganization method

jonatascastro12:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -135,11 +135,25 @@ type ListUsersOpts struct {
 }
 
 type CreateUserOpts struct {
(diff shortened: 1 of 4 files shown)


pkg/users/client.go
```diff
@@ -307,10 +316,11 @@ func (c *Client) AuthenticateUserWithPassword(ctx context.Context, opts Authenti
 
 // CreateUser create a new user with email password authentication.
```
AshGodfrey, line 317: think a comment for the wrong method snuck in here.

AshGodfrey (approved):

PR #227, opened 2023-08-08
User management: Add revoke session, verify session, and revoke all sessions

AshGodfrey:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -403,16 +403,17 @@ type VerifySessionResponse struct {
 }
 
 func (c *Client) VerifySession(ctx context.Context, opts VerifySessionOpts) (VerifySessionResponse, error) {
-	encodedForm, err := query.Values(opts)
-
+	data, err := json.Marshal(opts)
 	if err != nil {
 		return VerifySessionResponse{}, err
 	}
+
 	req, err := http.NewRequest(
 		http.MethodPost,
 		c.Endpoint+"/users/sessions/verify",
-		strings.NewReader(encodedForm.Encode()),
+		bytes.NewReader(data),  
 	)
+
 	if err != nil {
 		return VerifySessionResponse{}, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/users/client.go
```diff
@@ -392,3 +392,132 @@ func (c *Client) CreateUser(ctx context.Context, opts CreateUserOpts) (User, err
...
+		strings.NewReader(encodedForm.Encode()),
+	)
+	if err != nil {
+		return VerifySessionResponse{}, err
+	}
+
+	// Add headers and context to the request
+	req = req.WithContext(ctx)
+	req.Header.Set("User-Agent", "workos-go/"+workos.Version)
+	req.Header.Set("Authorization", "Bearer "+c.APIKey)
```
jonatascastro12, line 423: This one doesn't require `Authorization` too 😄

pkg/users/client.go
```diff
@@ -392,3 +392,132 @@ func (c *Client) CreateUser(ctx context.Context, opts CreateUserOpts) (User, err
...
+	var body VerifySessionResponse
+	dec := json.NewDecoder(res.Body)
+	err = dec.Decode(&body)
+
+	return body, err
+}
+
+type RevokeSessionOpts struct {
+	SessionToken string `json:"session_token"`
+	SessionID    string `json:"session_id"`
```
jonatascastro12, line 447:
```suggestion
	SessionToken string `json:"session_token,omitempty"`
	SessionID    string `json:"session_id,omitempty"`
```

pkg/users/client.go
AshGodfrey: Do users not need their API key for this one then?

jonatascastro12: Yes, they don't.

jonatascastro12 (approved):

PR #230, opened 2023-08-09
Update session interface to match api changes

AshGodfrey:
## Description

Matching Node changes: 
 https://github.com/workos/workos-node/pull/802

## Documentation

```
[ x] Yes
```

### DIFF
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -160,7 +160,7 @@ const (
 
 type UnauthorizedOrganizationReason struct {
 	Type                         string                        `json:"type"`
-	AllowedAuthenticationMethods []SessionAuthenticationMethod `json:"allowedAuthenticationMethods"`
+	AllowedAuthenticationMethods []SessionAuthenticationMethod `json:"allowed_authentication_methods"`
 }
 
 type UnauthorizedOrganization struct {
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -79,10 +79,6 @@ type User struct {
 	// The type of the User: `managed` or `unmanaged`
 	UserType UserType `json:"user_type"`
 
-	// List of the user's organization memberships. Unmanaged users can have zero or multiple memberships.
-	// Managed users have exactly one membership.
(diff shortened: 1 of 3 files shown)


pkg/users/client.go
```diff
@@ -134,32 +130,39 @@ type ListUsersOpts struct {
...
 
-type AddUserToOrganizationOpts struct {
-	User         string `json:"id"`
-	Organization string `json:"organization_id"`
+type Session struct {
+	ID                        string                     `json:"id"`
+	Token                     string                     `json:"token"`
+	CreatedAt                 string                     `json:"created_at"`
+	ExpiresAt                 string                     `json:"expires_at"`
+	AuthorizedOrganizations   []Organization             `json:"authorized_organizations"`
```
oliverzheng, line 145: There should be an `AuthorizedOrganization` struct, which has just 1 field `organization` within it, which has name and ID. It looks a little extra but it's by design.

AshGodfrey: Will make that update, thanks for confirming

pkg/users/client.go
```diff
@@ -134,32 +130,42 @@ type ListUsersOpts struct {
...
+const (
+	GoogleOauth    SessionAuthenticationMethod = "GoogleOauth"
+	MagicAuth      SessionAuthenticationMethod = "MagicAuth"
+	MicrosoftOauth SessionAuthenticationMethod = "MicrosoftOauth"
+	Password       SessionAuthenticationMethod = "Password"
+)
+
+type UnauthorizedOrganizationReason struct {
+	Type                         string                        `json:"type"`
+	AllowedAuthenticationMethods []SessionAuthenticationMethod `json:"allowedAuthenticationMethods"`
```
jonatascastro12, line 163:
```suggestion
	AllowedAuthenticationMethods []SessionAuthenticationMethod `json:"allowed_authentication_methods"`
```

jonatascastro12 (changes requested):
Need to change that camelCase value to snake case since it reflects the json field.

jonatascastro12 (approved):

PR #244, opened 2023-08-24
Remove MagicAuthChallenge from public interface

jonatascastro12:
## Description
- Replace `MagicAuthChallengeID` property by `User` (user ID) on `AuthenticateUserWithMagicAuthOpts`
- `SendMagicAuthCode` returns `User` instead of `MagicAuthChallengeID`

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -188,12 +188,12 @@ type AuthenticateUserWithCodeOpts struct {
 type MagicAuthChallengeID string
 
 type AuthenticateUserWithMagicAuthOpts struct {
-	ClientID             string               `json:"client_id"`
-	Code                 string               `json:"code"`
-	MagicAuthChallengeID MagicAuthChallengeID `json:"magic_auth_challenge_id"`
-	ExpiresIn            int                  `json:"expires_in,omitempty"`
-	IPAddress            string               `json:"ip_address,omitempty"`
-	UserAgent            string               `json:"user_agent,omitempty"`
+	ClientID  string `json:"client_id"`
+	Code      string `json:"code"`
+	User      string `json:"user_id"`
+	ExpiresIn int    `json:"expires_in,omitempty"`
+	IPAddress string `json:"ip_address,omitempty"`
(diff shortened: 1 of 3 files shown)


jonatascastro12:
Current dependencies on/for this PR:
* main
  * **PR #243** <a href="https://app.graphite.dev/github/pr/workos/workos-go/243" target="_blank">Graphite</a> 
    * **PR #244** <a href="https://app.graphite.dev/github/pr/workos/workos-go/244" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/workos/workos-go/244?utm_source=stack-comment).

pkg/users/client.go
```diff
@@ -819,15 +815,15 @@ func (c *Client) CompletePasswordReset(ctx context.Context, opts CompletePasswor
 }
 
 // SendMagicAuthCode creates a one-time Magic Auth code and emails it to the user.
-func (c *Client) SendMagicAuthCode(ctx context.Context, opts SendMagicAuthCodeOpts) (MagicAuthChallengeID, error) {
+func (c *Client) SendMagicAuthCode(ctx context.Context, opts SendMagicAuthCodeOpts) (User, error) {
```
AshGodfrey, line 818: I think this is where the downside of calling this "user" exists - I think we should evaluate if this should be "UserID" especially when we call it `ClientID`. 

I probably shouldn't make the final call or anything here, but it just seems confusing that we return a User object but ask people to pass in `id` as `User`

jonatascastro12: Yeah.. I see... but all the other methods have inputs without `Id` prefix. And also, you'll see the difference in the type (`User` vs `string`).

AshGodfrey: This isn't necessarily true in our other packages, so it's something I've been pushing for us to consider further. For example in `MFA` we have the `ID` suffix on all of the passed in values (and the naming convention changes there were the reason we moved from `v1` to `v2` of Go).

AshGodfrey (approved):

PR #255, opened 2023-09-06
Add `users.ListAuthFactor()` method

jonatascastro12:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -217,6 +217,10 @@ type EnrollAuthFactorOpts struct {
 	Type mfa.FactorType `json:"type"`
 }
 
+type ListAuthFactorsOpts struct {
+	User string
+}
+
 func NewClient(apiKey string) *Client {
 	return &Client{
 		APIKey:     apiKey,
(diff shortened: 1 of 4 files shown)


pkg/users/client.go
```diff
@@ -956,3 +960,41 @@ func (c *Client) EnrollAuthFactor(ctx context.Context, opts EnrollAuthFactorOpts
 
 	return body, err
 }
+
+// ListAuthFactors lists the available authentication factors for the user.
+func (c *Client) ListAuthFactors(ctx context.Context, opts ListAuthFactorsOpts) ([]mfa.Factor, error) {
```
AshGodfrey, line 965: Is the plan for these factors always to be the same (if so, I wonder if they should move to common in v3).

AshGodfrey (approved):

pkg/users/client.go
jonatascastro12: It makes sense. I'm following what they did on Node SDK.

PR #266, opened 2023-10-23
WIP: Add Invite Endpoints

AshGodfrey:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -238,6 +238,27 @@ type ListAuthFactorsResponse struct {
 	ListMetadata common.ListMetadata `json:"listMetadata"`
 }
 
+type InviteObject struct {
+	Object         string `json:"object"`
+	ID             string `json:"id"`
+	Email          string `json:"email"`
+	State          string `json:"state"`
+	AcceptedAt     string `json:"accepted_at,omitempty"`
+	RevokedAt      string `json:"revoked_at,omitempty"`
+	Token          string `json:"token"`
+	OrganizationID string `json:"organization_id,omitempty"`
+	ExpiresAt      string `json:"expires_at"`
+	CreatedAt      string `json:"created_at"`
+	UpdatedAt      string `json:"updated_at"`
+}
+
+type CreateInvitationOpts struct {
+	Email          string `json:"email"`
+	OrganizationID string `json:"organization_id,omitempty"`
+	ExpiresInDays  int    `json:"expires_in_days,omitempty"`
+	InviterUserID  string `json:"inviter_user_id,omitempty"`
+}
+
 func NewClient(apiKey string) *Client {
 	return &Client{
 		APIKey:     apiKey,
@@ -1061,6 +1082,44 @@ func (c *Client) ListAuthFactors(ctx context.Context, opts ListAuthFactorsOpts) 
 	}
 
 	var body ListAuthFactorsResponse
+	dec := json.NewDecoder(res.Body)
+	err = dec.Decode(&body)
+
+	return body, err
+}
+
+func (c *Client) CreateInvitation(ctx context.Context, opts CreateInvitationOpts) (InviteObject, error) {
+	endpoint := fmt.Sprintf("%s/users/invites", c.Endpoint)
+
+	data, err := json.Marshal(opts)
+	if err != nil {
+		return InviteObject{}, err
+	}
+
+	req, err := http.NewRequest(
+		http.MethodPost,
+		endpoint,
+		bytes.NewBuffer(data),
+	)
+	if err != nil {
+		return InviteObject{}, err
+	}
+	req = req.WithContext(ctx)
+	req.Header.Set("User-Agent", "workos-go/"+workos.Version)
+	req.Header.Set("Authorization", "Bearer "+c.APIKey)
+	req.Header.Set("Content-Type", "application/json")
+
+	res, err := c.HTTPClient.Do(req)
+	if err != nil {
+		return InviteObject{}, err
+	}
+	defer res.Body.Close()
+
+	if err = workos_errors.TryGetHTTPError(res); err != nil {
+		return InviteObject{}, err
+	}
+
+	var body InviteObject
 	dec := json.NewDecoder(res.Body)
 	err = dec.Decode(&body)
 
--- a/pkg/users/client.go
+++ b/pkg/users/client.go
@@ -26,6 +26,17 @@ const (
 	Desc Order = "desc"
 )
 
+// TokenState represents the state of a token.
+type TokenState string
+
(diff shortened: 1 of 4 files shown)


pkg/users/client.go
```diff
@@ -238,6 +238,44 @@ type ListAuthFactorsResponse struct {
 	ListMetadata common.ListMetadata `json:"listMetadata"`
 }
 
+type InviteObject struct {
+	Object         string `json:"object"`
```
amygdalama, line 242:
```suggestion
type Invite struct {
```
small naming change + looks like we omit the object field for other object types.

pkg/users/client.go
```diff
@@ -238,6 +238,44 @@ type ListAuthFactorsResponse struct {
...
+	ID             string `json:"id"`
+	Email          string `json:"email"`
+	State          string `json:"state"`
```
amygdalama, line 245: enumerate the different `state`s?
* pending
* accepted
* expired
* revoked

(not sure how to do that in go, hopefully we have some existing examples to follow...)

pkg/users/client.go
```diff
@@ -238,6 +238,44 @@ type ListAuthFactorsResponse struct {
...
+	AcceptedAt     string `json:"accepted_at,omitempty"`
+	RevokedAt      string `json:"revoked_at,omitempty"`
+	Token          string `json:"token"`
+	OrganizationID string `json:"organization_id,omitempty"`
+	ExpiresAt      string `json:"expires_at"`
+	CreatedAt      string `json:"created_at"`
+	UpdatedAt      string `json:"updated_at"`
+}
+
+type CreateInvitationOpts struct {
```
amygdalama, line 255: we call these "invite" / "invites" in our API so i think we should be consistent w/ that in the SDKs

pkg/users/client.go
```diff
@@ -238,6 +238,44 @@ type ListAuthFactorsResponse struct {
...
+type RevokeInvitationOpts struct {
+	InviteID string
+}
+
+type ListInvitationsOpts struct {
+	OrganizationID string `json:"organization_id,omitempty"`
+	Email          string `json:"email,omitempty"`
+}
+
+type GetInvitationByIDOpts struct {
```
amygdalama, line 271:
```suggestion
type GetInviteOpts struct {
```

looks like the corresponding types for other endpoints don't have a "by ID" suffix (like `GetUserOpts`). let's match that convention; it also matches the API.

pkg/users/client.go
```diff
@@ -1066,3 +1104,174 @@ func (c *Client) ListAuthFactors(ctx context.Context, opts ListAuthFactorsOpts)
...
+
+	var body InviteObject
+	dec := json.NewDecoder(res.Body)
+	err = dec.Decode(&body)
+
+	return body, err
+}
+
+// ListInvitations gets a list of all of your existing invites matching the criteria specified.
+func (c *Client) ListInvitations(ctx context.Context, opts ListInvitationsOpts) ([]InviteObject, error) {
```
amygdalama, line 1176: what about the pagination part of the response?

amygdalama (commented):
i only reviewed the types + responses (deferring correctness / go conventions to someone in the go group)

thank you for picking this up!! 🙏

pkg/users/client.go
AshGodfrey: No worries - easy enough, I'll add these :)

pkg/users/client.go
AshGodfrey: How does this work when we also have getInvitationByToken? It seems like token and ID should have separate options to pass in? Or are we using a "GetInviteOpts" with a token and ID for both?

AshGodfrey: Or is one `GetInviteOpts` and the other `GetInviteByTokenOpts`?

AshGodfrey: We usually use the same name as the function. So if it's "`GetInvite`" for ID and "`GetInviteByToken`" for token this makes sense. However if we want `GetInviteByID`, structurally we probably want to go with `GetInviteByIDOpts`

amygdalama: yeah i was thinking two separate functions `GetInvite` and `GetInviteByToken` (which matches the API implementation). i think as a default we should match the API implementation to keep it straightforward, and then we can make exceptions in cases where the API is inconsistent or not idiomatic. do you think this is one of those cases?

AshGodfrey: Cool as long as it's consistent, I'm 100% for it - will push these changes today.

pkg/users/client.go
```diff
@@ -26,6 +26,17 @@ const (
 	Desc Order = "desc"
 )
 
+// TokenState represents the state of a token.
+type TokenState string
```
amygdalama, line 30: call this `InviteState` instead pls!

pkg/users/client.go
```diff
@@ -238,6 +249,65 @@ type ListAuthFactorsResponse struct {
...
+	// Pagination cursor to receive records before a provided User ID.
+	Before string `url:"before,omitempty"`
+
+	// Pagination cursor to receive records after a provided User ID.
+	After string `url:"after,omitempty"`
+}
+
+type GetInviteOpts struct {
+	InviteID string
+}
```
amygdalama, line 305:
```suggestion
type GetInviteOpts struct {
        // Invite unique identifier
	Invite string `json:"id"`
}
```
looks like in other `Get*Opts` cases we use the object name here instead, so let's follow that convention:
https://github.com/workos/workos-go/blob/c831fd777f3fa50f092ab4cb880bb768f33a3c14/pkg/users/client.go#L86-L90

(sorry i didn't catch this before)

amygdalama: idk if `json:"id"` is correct btw i just copy/pasta-ed from the user case

amygdalama: and i suggest using that same convention for `RevokeInviteOpts`

AshGodfrey: yeah this is a contested naming convention at times, sometimes we call it `User` other times we call it `UserID` or just `ID` (this particular package is going with the `User` style convention but we've debated it in slack). I'll go ahead and call it `User` style for package consistency

AshGodfrey: Also it doesn't need the JSON string here because it's being passed in in the URL.

AshGodfrey: JSON tags are irrelevant in the opts - just the response where there's JSON encoding/decoding.

amygdalama (approved):
names + types look good to me!

will defer to someone from the go group on implementation specifics

amygdalama:
hold up on this; we're making some last-minute changes to the API

PR #278, opened 2023-11-20
Add Organization Memberships API

amygdalama:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -38,8 +38,14 @@ type Organization struct {
 
 // OrganizationMembership contains data about a particular OrganizationMembership.
 type OrganizationMembership struct {
-	// Contains the ID and name of the associated Organization.
-	Organization Organization `json:"organization"`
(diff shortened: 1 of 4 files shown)


pkg/usermanagement/client.go
```diff
@@ -932,3 +984,161 @@ func (c *Client) ListAuthFactors(ctx context.Context, opts ListAuthFactorsOpts)
...
+	}
+	defer res.Body.Close()
+
+	if err = workos_errors.TryGetHTTPError(res); err != nil {
+		return OrganizationMembership{}, err
+	}
+
+	var body OrganizationMembership
+	dec := json.NewDecoder(res.Body)
+	err = dec.Decode(&body)
```
jbvolta, line 1021: could be safer to 
```if err != nil
  return return OrganizationMembership{}, err

return body ```

jbvolta: but better to stay consistent with all other impl

jbvolta (approved):

PR #285, opened 2023-11-30
Add AuthenticateWithEmailVerificationCode

amygdalama:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -176,6 +176,14 @@ type AuthenticateWithTOTPOpts struct {
 	AuthenticationChallengeID  string `json:"authentication_challenge_id"`
 }
 
+type AuthenticateWithEmailVerificationCodeOpts struct {
+	ClientID                   string `json:"client_id"`
+	Code                       string `json:"code"`
+	PendingAuthenticationToken string `json:"pending_authentication_token"`
+	IPAddress                  string `json:"ip_address,omitempty"`
+	UserAgent                  string `json:"user_agent,omitempty"`
+}
+
 type AuthenticateResponse struct {
 	User User `json:"user"`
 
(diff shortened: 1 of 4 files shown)


pkg/usermanagement/client.go
```diff
@@ -176,6 +176,14 @@ type AuthenticateWithTOTPOpts struct {
 	AuthenticationChallengeID  string `json:"authentication_challenge_id"`
 }
 
+type AuthenticateWithEmailVerificationCodeOpts struct {
```
danerwilliams, line 179: could we extend this from a base authenticate type like we do in node?
https://github.com/workos/workos-node/blob/7f9024e45bc8d0662ec975ea9517a91f4c3625ee/src/user-management/interfaces/authenticate-with-email-verification-options.interface.ts#L6-L10
https://github.com/workos/workos-node/blob/7f9024e45bc8d0662ec975ea9517a91f4c3625ee/src/user-management/interfaces/authenticate-with-options-base.interface.ts#L1-L5

I believe we can do something like this:

```go
type AuthenticateWithOptionsBase struct {
    ClientID  string `json:"client_id"`
    IPAddress string `json:"ip_address,omitempty"`
    UserAgent string `json:"user_agent,omitempty"`
}
```

```go
type AuthenticateWithEmailVerificationCodeOpts struct {
    AuthenticateWithOptionsBase
    Code                       string `json:"code"`
    PendingAuthenticationToken string `json:"pending_authentication_token"`
}
```

danerwilliams (approved):

pkg/usermanagement/client.go
amygdalama: i would like to minimize refactoring until we get the critical pieces done since we're aiming for an end-of-the-week release...

amygdalama: (unless it's something user-facing)

PR #297, opened 2023-12-02
Fix TOTP grant type

amygdalama:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -840,7 +840,7 @@ func (c *Client) AuthenticateWithTOTP(ctx context.Context, opts AuthenticateWith
 	}{
 		AuthenticateWithTOTPOpts: opts,
 		ClientSecret:             c.APIKey,
-		GrantType:                "workos:oauth:grant-type:mfa-totp",
+		GrantType:                "urn:workos:oauth:grant-type:mfa-totp",
 	}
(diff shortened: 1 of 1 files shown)


pkg/usermanagement/client.go
```diff
@@ -840,7 +840,7 @@ func (c *Client) AuthenticateWithTOTP(ctx context.Context, opts AuthenticateWith
 	}{
 		AuthenticateWithTOTPOpts: opts,
 		ClientSecret:             c.APIKey,
-		GrantType:                "workos:oauth:grant-type:mfa-totp",
+		GrantType:                "urn:workos:oauth:grant-type:mfa-totp",
```
jbvolta, line 843: Is it possible to add that to unit test ?

jbvolta (approved):

pkg/usermanagement/client.go
amygdalama: hmm, i'm not sure how to do that in Go... is there a way to assert a method was called with an argument w/ a particular value? or would we need to install a package for that?

here's what the unit test currently looks like https://github.com/workos/workos-go/blob/cd25f949c72b6db1f80e5e0ad20c5f0258105eb7/pkg/usermanagement/client_test.go#L796-L811

PR #304, opened 2024-02-16
Update README.md

klausbreyer:
Removed dots to make the command copy & pastable.


klausbreyer:
ok, I know I got why the ellipsis is there. you really should explain that you need to install every package on its own in all your docs, please :)

PR #308, opened 2024-03-14
Add role to OrganizationMembership and new PUT OrganizationMembership endpoint

ameesha:
## Description
New "role" parameter which is used in the User Management OrganizationMemberships API

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -71,6 +71,11 @@ const (
 	PendingOrganizationMembership OrganizationMembershipStatus = "pending"
 )
 
+type RoleResponse struct {
+	// The slug of the role
+	Slug string `json:"slug"`
+}
+
 // OrganizationMembership contains data about a particular OrganizationMembership.
 type OrganizationMembership struct {
 	// The Organization Membership's unique identifier.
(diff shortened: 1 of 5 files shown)


pkg/usermanagement/client.go
```diff
@@ -347,6 +355,10 @@ type CreateOrganizationMembershipOpts struct {
 
 	// The ID of the Organization in which to add the User as a member.
 	OrganizationID string `json:"organization_id"`
+
+	// The slug of the Role in which to this membership. If not RoleSlug is given, the default role will be granted.
```
mthadley, line 359:
```suggestion
	// The slug of the Role in which to this membership. If no RoleSlug is given, the default role will be granted.
```

pkg/usermanagement/client.go
```diff
@@ -347,6 +355,10 @@ type CreateOrganizationMembershipOpts struct {
...
+	// OPTIONAL
+	RoleSlug string `json:"role_slug"`
```
mthadley, line 361: I'm guessing empty string is not a valid slug, so I think we probably want to add `omitempty` so that it isn't serialized into the request body:

```suggestion
	RoleSlug string `json:"role_slug,omitempty"`
```

mthadley (approved):

PR #309, opened 2024-03-14
Add session endpoints

cmatheson:
## Description

Adds the types/endpoints necessary for dealing with sessions.

## Documentation

```
[X] Yes
```

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -419,6 +419,10 @@ type RevokeInvitationOpts struct {
 	Invitation string
 }
 
+type RevokeSessionOpts struct {
+	SessionID string `json:"session_id"`
+}
+
 func NewClient(apiKey string) *Client {
 	return &Client{
 		APIKey:     apiKey,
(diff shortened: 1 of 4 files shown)


pkg/usermanagement/client.go
```diff
@@ -798,6 +825,58 @@ func (c *Client) AuthenticateWithCode(ctx context.Context, opts AuthenticateWith
...
+		c.Endpoint+"/user_management/authenticate",
+		bytes.NewBuffer(jsonData),
+	)
+
+	if err != nil {
+		return RefreshAuthenticationResponse{}, err
+	}
+
+	// Add headers and context to the request
+	req = req.WithContext(ctx)
```
mthadley, line 857: I know this is just following existing patterns, and if we refactor it should be done separately, but I wonder why we don't call [`http.NewRequestWithContext`](https://pkg.go.dev/net/http#NewRequestWithContext) initially, rather than separate reassigning `req` later like this.

mthadley (approved):

cmatheson:
bors merge

cmatheson:
bors cancel

PR #325, opened 2024-04-09
Make events parameter required for ListEvents

mattgd:
## Description
Make events parameter required for ListEvents.

## Documentation

```
[x] Yes
```

https://github.com/workos/workos/pull/26139

### DIFF
--- a/internal/workos/workos.go
+++ b/internal/workos/workos.go
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v4.6.0"
+	Version = "v4.5.0"
 )
--- a/internal/workos/workos.go
+++ b/internal/workos/workos.go
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v4.5.0"
+	Version = "v4.6.0"
 )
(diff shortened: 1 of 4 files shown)


PaulAsjes (approved):

pkg/events/client.go
```diff
@@ -89,7 +89,7 @@ type Event struct {
 // ListEventsOpts contains the options to request provisioned Events.
 type ListEventsOpts struct {
 	// Filter to only return Events of particular types.
-	Events []string `url:"events,omitempty"`
+	Events []string `url:"events"`
```
mattgd, line 92: @PaulAsjes While this change is breaking in the API (mitigated by feature flag for affected customers), I don't consider this is a breaking change in the Go SDK, so a minor bump seems appropriate. Wanted to double check to see if you had an opinion though.

PaulAsjes: Agreed, minor bump seems fine!

internal/workos/workos.go
```diff
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v4.5.0"
+	Version = "v4.6.0"
```
mthadley, line 5: Can we bump the version in a separate PR? Like we've done [here](https://github.com/workos/workos-go/pull/321) and in earlier releases.

There are other unreleased changes on `main`, so it makes the commit history easier to view in regards to what was included in each version bump.

mattgd: Yep for sure, will split.

mthadley (approved):

PR #327, opened 2024-04-19
Add `domains_data` to Organization API and deprecate `domains`

oliverzheng:
## Documentation

```
[ ] Yes
```


pkg/organizations/client.go
```diff
@@ -134,9 +150,12 @@ type CreateOrganizationOpts struct {
 	// outside of the Organization's configured User Email Domains.
 	AllowProfilesOutsideOrganization bool `json:"allow_profiles_outside_organization"`
 
-	// Domains of the Organization.
+	// [Deprecated] Domains of the Organization. Use DomainData instead.
```
mthadley, line 153: Looks like this should be formatted as `Deprecated:` instead:

https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go

mthadley (approved):

PR #317, opened 2024-03-26
Add authentication events

LauraBeatris:
## Documentation

Does this require changes to the WorkOS Docs? E.g. the [API Reference](https://workos.com/docs/reference) or code snippets need updates.

```
[ ] Yes
```

If yes, link a related docs PR and add a docs maintainer as a reviewer. Their approval is required.

PR #336, opened 2024-05-17
Revert "Add authentication events"

mattgd:
Reverts workos/workos-go#317 as this was merged on accident. Will recreate for a future release.


ameesha (approved):
can you add why you're reverting as well

mattgd:
> can you add why you're reverting as well

Good call, added.

PR #350, opened 2024-06-18
Fix struct tag

chandlerroth:
## Description

Found a little SDK bug here. This is breaking requests with the `GET /user_management/invitations` endpoint. Currently the SDK does not have the correct struct tags for the `ListInvitationsOpts` struct. 

Before:
Screenshot 2024-06-18 at 16 24 15

After:
Screenshot 2024-06-18 at 16 25 41

## Documentation

Does this require changes to the WorkOS Docs? 

> No.

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -512,9 +512,9 @@ type ListInvitationsResponse struct {
 }
 
 type ListInvitationsOpts struct {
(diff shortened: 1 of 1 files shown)


PaulAsjes (approved):

PaulAsjes:
Thanks for this!

PR #353, opened 2024-07-10
Add FGA client to support FGA endpoints

stanleyphu:
## Description
This PR creates a new FGA package and client that adds support for FGA endpoints, including `resources`, `warrants`, `check` and `query`.

## Documentation

```
[x] Yes
```

https://github.com/workos/workos/pull/28797

### DIFF
--- a/pkg/fga/client.go
+++ b/pkg/fga/client.go
@@ -139,6 +139,45 @@ type DeleteObjectOpts struct {
 	ObjectId string
 }
 
+// Object types
+type ObjectType struct {
+	// Unique string ID of the object type.
+	Type string `json:"type"`
+
+	// Set of relationships that subjects can have on objects of this type.
+	Relations map[string]interface{} `json:"relations"`
+}
+
+type ListObjectTypesOpts struct {
+	// Maximum number of records to return.
+	Limit int `url:"limit,omitempty"`
+
+	// The order in which to paginate records.
+	Order Order `url:"order,omitempty"`
+
+	// Pagination cursor to receive records before a provided ObjectType ID.
+	Before string `url:"before,omitempty"`
+
+	// Pagination cursor to receive records after a provided ObjectType ID.
+	After string `url:"after,omitempty"`
+}
+
+type ListObjectTypesResponse struct {
+	// List of Object Types.
+	Data []ObjectType `json:"data"`
+
+	// Cursor pagination options.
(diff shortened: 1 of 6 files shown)


pkg/fga/client.go
```diff
@@ -0,0 +1,942 @@
...
+	}
+
+	var jsonCtx []byte
+	if opts.Context != nil {
+		jsonCtx, err = json.Marshal(opts.Context)
+		if err != nil {
+			return QueryResponse{}, err
+		}
+	}
+	queryUrlOpts := QueryUrlOpts{
```
kkajla12, line 912: Rather than defining this inline struct to accommodate context being a `string` vs a `Context` (so that the conversion to url params is done correctly), we should define an `EncodeValues` method on the `QueryOpts` struct (so it implements the [`query.Encoder`](https://pkg.go.dev/github.com/google/go-querystring/query#Encoder) interface) which converts the `Context` to a `string` upon conversion to url values.

kkajla12 (approved):

kkajla12 (approved):

pkg/fga/client.go
```diff
@@ -0,0 +1,926 @@
...
+	SubjectRelation string `url:"subject_relation,omitempty"`
+
+	// Maximum number of records to return.
+	Limit int `url:"limit,omitempty"`
+
+	// Pagination cursor to receive records after a provided Warrant ID.
+	After string `url:"after,omitempty"`
+
+	// Optional token to specify desired read consistency
+	WarrantToken string `url:"-"`
```
mthadley, line 244: Is it intentional to completely ignore this `WarrantToken` field when encoding this struct as query params?

mthadley: Ah, I see why now. It ends up being sent as a request header. 👍

pkg/organizations/client.go
```diff
@@ -163,7 +163,7 @@ type CreateOrganizationOpts struct {
 	DomainData []OrganizationDomainData `json:"domain_data"`
 
 	// Optional unique identifier to ensure idempotency
-	IdempotencyKey string `json:"idempotency_iey,omitempty"`
+	IdempotencyKey string `json:"idempotency_key,omitempty"`
```
mthadley, line 166: 🙈

mthadley (approved):

PR #356, opened 2024-08-06
Add retryable http client and use in FGA module

stanleyphu:
## Description

- Adds `httpretryable.HttpClient` which implements the `.Do` method for making requests with built-in retry logic using exponential backoff.

- Replaces the http client in the FGA module to use the retryable http client

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/retryablehttp/client.go
+++ b/pkg/retryablehttp/client.go
@@ -1,0 +1,107 @@
+package retryablehttp
+
+import (
+	"io"
+	"math"
+	"math/rand"
+	"net/http"
+	"time"
+)
+
+const MaxRetryAttempts = 3
+const MinimumDelay = 500
+const MinimumDelayDuration = 500 * time.Millisecond
+const MaximumDelayDuration = 5 * time.Second
+const RandomizationFactor = 0.5
+const BackoffMultiplier = 1.5
+
+type HttpClient struct {
+	http.Client
+}
+
+func (client *HttpClient) Do(req *http.Request) (*http.Response, error) {
+	var res *http.Response
+	var err error
+	for retry := 0; ; {
+		// Reset the request body for each retry
+		if req.Body != nil {
+			body, err := req.GetBody()
+			if err != nil {
+				client.CloseIdleConnections()
+				return res, err
+			}
+			if c, ok := body.(io.ReadCloser); ok {
+				req.Body = c
+			} else {
+				req.Body = io.NopCloser(body)
+			}
+		}
+
+		res, err = client.Client.Do(req)
+		if err != nil {
+			break
+		}
+
+		shouldRetry := client.shouldRetry(req, res, err, retry)
+
+		if !shouldRetry {
+			break
+		}
+
+		sleepTime := client.sleepTime(retry)
+		retry++
+
+		time.Sleep(sleepTime)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	return res, nil
+}
+
+func (client *HttpClient) shouldRetry(req *http.Request, resp *http.Response, err error, retryAttempt int) bool {
+	if retryAttempt >= MaxRetryAttempts {
+		return false
+	}
+
+	if req.Context() != nil && req.Context().Err() != nil {
+		return false
+	}
+
+	if err != nil {
+		return true
+	}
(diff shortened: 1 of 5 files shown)


pkg/retryablehttp/client.go
```diff
@@ -0,0 +1,107 @@
...
+		shouldRetry := client.shouldRetry(req, res, err, retry)
+
+		if !shouldRetry {
+			break
+		}
+
+		sleepTime := client.sleepTime(retry)
+		retry++
+
+		time.Sleep(sleepTime)
```
mthadley, line 54: Should this be `select`ing on the `Context` from the request in order to not block the caller while it waits to retry?

An example of the concurrency pattern I'm thinking of (just pulled randomly from search, not proposing we use this package):

https://github.com/sethvargo/go-retry/blob/9bf6450b891a2dfa8064b27d3b876fcdb83be28e/retry.go#L86-L93

kkajla12: Won't that effectively make all of our FGA methods asynchronous? We could go down that route, but that would mean we would need to return a done channel from all FGA methods that the caller (customer application) could listen to for the eventual result. By default, the Go standard library's [`http.Server`](https://pkg.go.dev/net/http#Server) runs every request on a separate goroutine (thread), which means that we'll only be blocking the customer application's goroutine handling that particular request. I believe this is the desired/optimal behavior. In case customers would like to run our SDK methods asynchronously, they can choose to run them in a separate goroutine if they'd like.
```go
go func() {
   res, err := fga.WriteWarrants(...)
   if err != nil {
      ...
   }
   ...
}()
```

Also note that [`time.Sleep` still yields to other goroutines](https://stackoverflow.com/a/34668153), so that's good.

mthadley: > Won't that effectively make all of our FGA methods asynchronous?

No, or at least AFAICT it shouldn't. Looking at that example I previously linked, the result of the operation wrapped in the try logic is still provided to the caller as a normal return value; not by sending it through a channel.

https://github.com/sethvargo/go-retry/blob/9bf6450b891a2dfa8064b27d3b876fcdb83be28e/retry.go#L65

```go
v, err := f(ctx)
if err == nil {
	return v, nil
}
```

pkg/retryablehttp/client.go
```diff
@@ -0,0 +1,107 @@
+package retryablehttp
```
kkajla12, line 1: The ideal name for this package is probably `http` IMO. I know the conflict with `net/http` could be annoying, but we can alias our package as `wohttp` or something. We should also update the main struct from `HttpClient` to `HttpClientWithRetry` in this case.

pkg/retryablehttp/client.go
mthadley: Oh, I just reread my wording:

> in order to not block the caller while it waits to retry?

Yeah, that totally sounds like I was suggesting it should be asynchronous. My bad.

I was more thinking about how the goroutine might be "blocking" say a server or background job, from shutting down/cancelling, because the method was sleeping.

kkajla12: Oh, it looks like the example you linked to is using the `select` statement to check for `ctx.Done()` to support proper request cancellation (unrelated to blocking the current thread). We should probably do this as well, in which case we'd want to accept the current context as the first argument to the `Do` method @stanleyphu.

mthadley: > Oh, it looks like the example you linked to is using the select statement to check for ctx.Done() to support proper request cancellation (unrelated to blocking the current thread).

Yes, exactly! I know of the pattern (since the rest of the SDK was written with it in mind), but was grasping for the right Go terminology. Thanks for clarifying that.

stanleyphu: Updated the `Do` method to create a timer which then allows us to do a `select` to allow the timer to complete or the context to be cancelled.

pkg/retryablehttp/client.go
kkajla12: After discussing, we decided to keep this package named `retryablehttp` and keep the struct named `HttpClient`

kkajla12 (approved):

PR #364, opened 2024-08-27
[FGA-98] Add ConvertSchemaToResourceTypes and ConvertResourceTypesToSchema to FGA module

atainter:
## Description
Adds methods to convert new schema language to resource types JSON

### DIFF
--- a/internal/workos/workos.go
+++ b/internal/workos/workos.go
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v4.21.0"
+	Version = "v4.20.0"
 )
--- a/internal/workos/workos.go
+++ b/internal/workos/workos.go
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v4.20.0"
+	Version = "v4.21.0"
 )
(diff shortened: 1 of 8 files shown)


internal/workos/workos.go
```diff
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
-	Version = "v4.20.0"
+	Version = "v4.21.0"
```
kkajla12, line 5: I think this version update is typically done in a separate PR

kkajla12 (approved):

pkg/workos_errors/http.go
```diff
@@ -43,6 +43,34 @@ func isJsonResponse(r *http.Response) bool {
...
+	// Try unmarshaling as []ErrorWithMessage
+	var structErrors []ErrorWithMessage
+	if err := json.Unmarshal(data, &structErrors); err == nil {
+		for _, se := range structErrors {
+			*e = append(*e, se.Message)
+		}
+		return nil
+	}
+
+	return fmt.Errorf("Errors field is not a valid format")
```
kkajla12, line 71: Since we don't need to format anything in the error, we can `return errors.New("Errors field is not a valid format")` here instead.

kkajla12 (approved):

PR #369, opened 2024-10-15
Fix typo on DirectoryGroupUserRemoved event

awolfden:
## Description
Fix typo on the DirectoryGroupUserRemoved event 

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/events/client.go
+++ b/pkg/events/client.go
@@ -35,7 +35,7 @@ const (
 	DirectoryGroupUpdated     = "dsync.group.updated"
 	DirectoryGroupDeleted     = "dsync.group.deleted"
 	DirectoryGroupUserAdded   = "dsync.group.user_added"
-	DirectroyGroupUserRemoved = "dsync.group.user_removed"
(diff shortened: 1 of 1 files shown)


pkg/events/client.go
```diff
@@ -35,7 +35,7 @@ const (
 	DirectoryGroupUpdated     = "dsync.group.updated"
 	DirectoryGroupDeleted     = "dsync.group.deleted"
 	DirectoryGroupUserAdded   = "dsync.group.user_added"
-	DirectroyGroupUserRemoved = "dsync.group.user_removed"
+	DirectoryGroupUserRemoved = "dsync.group.user_removed"
```
mthadley, line 38: I think this would technically be a breaking change, but we can keep the old constant around and leave a deprecation comment:

```suggestion
	DirectoryGroupUserRemoved = "dsync.group.user_removed"
	DirectroyGroupUserRemoved = "dsync.group.user_removed" // Deprecated: use DirectoryGroupUserRemoved instead
```

Then we can remove it next time we do a major release.

mthadley (approved):

PR #376, opened 2024-11-08
Add new organization domain fields

amygdalama:
## Description
Adds fields introduced with the domain verification product (`State`, `VerificationStrategy`, `VerificationToken`), and `OrganizationID`, which is useful for parsing organization domain events.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/organizations/client.go
+++ b/pkg/organizations/client.go
@@ -84,6 +84,9 @@ type OrganizationDomain struct {
 	// The domain value
 	Domain string `json:"domain"`
 
+	// The Organization's unique identifier.
+	OrganizationID string `json:"organization_id"`
+
 	// Verification state of the domain.
 	State OrganizationDomainState `json:"state"`
 
--- a/pkg/organizations/client.go
+++ b/pkg/organizations/client.go
@@ -63,10 +63,10 @@ func (c *Client) init() {
 type OrganizationDomainState string
 
 const (
-	// Pending        OrganizationDomainState = "pending"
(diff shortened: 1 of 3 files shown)


pkg/organizations/client.go
```diff
@@ -60,13 +60,41 @@ func (c *Client) init() {
...
+	OrganizationID string `json:"organization_id"`
+
+	// Verification state of the domain.
+	State OrganizationDomainState `json:"state"`
+
+	// Strategy used to verify the domain.
+	VerificationStrategy OrganizationDomainVerificationStrategy `json:"verification_strategy,omitempty"`
+
+	// Token used for DNS verification.
+	VerificationToken string `json:"verification_token,omitempty"`
```
mattgd, line 97: Based on your thread, is your ultimate goal to have the domain data on the `organization_domain` events only, but for now they also reside on the `organization` events?

mattgd (approved):
Looks good, just had a high level question.

pkg/organizations/client.go
amygdalama: oh no, I meant that the `OrganizationID` specifically is mostly useful in the context of organization domain events (since in organization events or the organization API you have the ID already from the parent organization object). it's included by the API in all cases though.

for the other fields, i don't expect `VerificationStrategy` or `VerificationToken` to be widely used, but i'm including them for completeness. `State` on the other hand is generally useful in all cases where this struct would be used

PR #378, opened 2024-11-14
Add widgets client and get token API support

mattgd:
## Description
Add widgets client and get token API support.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/widgets/README.md
+++ b/pkg/widgets/README.md
@@ -1,0 +1,15 @@
+# widgets
+
+[![Go Report Card](https://img.shields.io/badge/dev-reference-007d9c?logo=go&logoColor=white&style=flat)](https://pkg.go.dev/github.com/workos/workos-go/v4/pkg/widgets)
+
+A Go package to make requests to the WorkOS Widgets API.
+
+## Install
+
+```sh
+go get -u github.com/workos/workos-go/v4/pkg/widgets
+```
+
+## How it works
+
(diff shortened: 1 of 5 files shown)


pkg/widgets/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Install
+
+```sh
+go get -u github.com/workos/workos-go/v4/pkg/widgets
+```
+
+## How it works
+
+See the [Widgets integration guide](https://workos.com/docs/widgets/guide).
```
mattgd, line 15: This page doesn't exist yet and should be updated before merging.

pkg/widgets/client.go
```diff
@@ -0,0 +1,115 @@
...
+	once sync.Once
+}
+
+func (c *Client) init() {
+	if c.HTTPClient == nil {
+		c.HTTPClient = &http.Client{Timeout: 10 * time.Second}
+	}
+
+	if c.Endpoint == "" {
+		c.Endpoint = "https://api.workos.com"
```
PaulAsjes, line 44: This should ideally be configurable. You can't currently change the API host in the golang SDK, but it's something we should add and that should reflect here too.

mattgd: Great point. I also feel like it's less than ideal that we have to redefine this common client logic for every module. Wondering if it should be refactored into a shared module.

PR #384, opened 2024-12-13
Add `NewClient` method to `organizations` package

mattt:
## Description

I was integrating WorkOS into my project and noticed that the `usermanagement` package had a `NewClient` method but the `organizations` package didn't. This PR adds that.

## Documentation

Does this require changes to the WorkOS Docs?  **No**

The examples in [the API Reference](https://workos.com/docs/reference) all use the default client and `SetAPIKey`.

### DIFF
--- a/pkg/organizations/client.go
+++ b/pkg/organizations/client.go
@@ -46,6 +46,15 @@ type Client struct {
 	once sync.Once
 }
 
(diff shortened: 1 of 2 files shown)


gjtorikian:
Thanks for your contribution! This PR was superseded by the new v7 SDK which adds a unified client.

PR #401, opened 2025-03-03
Add email standard attribute to DirectoryUser and mark deprecated standard attributes

mattgd:
## Description
Add email standard attribute to DirectoryUser and mark deprecated standard attributes.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/directorysync/client.go
+++ b/pkg/directorysync/client.go
@@ -99,10 +99,19 @@ type User struct {
 	// The identifier for the Organization in which the Directory resides.
 	OrganizationID string `json:"organization_id"`
 
+	// The User's primary email
+	Email string `json:"email"`
+
 	// The User's username.
+	// Deprecated: Will be removed in a future major version. Enable the `username` custom attribute
+	// in dashboard and pull from customAttributes instead. See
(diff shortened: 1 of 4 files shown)


pkg/directorysync/client.go
```diff
@@ -99,10 +99,19 @@ type User struct {
 	// The identifier for the Organization in which the Directory resides.
 	OrganizationID string `json:"organization_id"`
 
+	// The User's primary email
+	Email string `json:"email"`
+
 	// The User's username.
+	// Deprecated: Will be removed in a future major version. Enable the `username` custom attribute
```
nave91, line 106: Do you know if editors surface this up to the developers when using these attributes?

I see that [this is the convention](https://go.dev/wiki/Deprecated). I'm curious if this works with attributes in a struct.

nave91 (approved):

pkg/directorysync/client.go
mattgd: According to the doc you linked, this should show up as deprecated in editors, but in my editor all I see is the comment, no strikethrough or yellow squiggles or anything else.

jonatascastro12 (approved):
lgtm

PR #412, opened 2025-03-28
Add OrganizationDomains package

ajworkos:
## Description

Adds the [OrganizationDomains](https://workos.com/docs/reference/domain-verification/) package

## Documentation

Does this require changes to the WorkOS Docs? E.g. the [API Reference](https://workos.com/docs/reference) or code snippets need updates.

```
[ ] Yes
```

If yes, link a related docs PR and add a docs maintainer as a reviewer. Their approval is required.

PR #415, opened 2025-04-01
Alias organization domain types

ajworkos:
## Description

The [PR](https://github.com/workos/workos-go/pull/412) that added the `organization_domains` package moved the domain types to `organization_domains` from `organizations`. To prevent a major version bump and a breaking change, this PR aliases the moved types so they are still available via the `organizations` package

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/organizations/client.go
+++ b/pkg/organizations/client.go
@@ -64,7 +64,19 @@ func (c *Client) init() {
 
 type OrganizationDomainState = organization_domains.OrganizationDomainState
 
+const (
+	OrganizationDomainPending        OrganizationDomainState = organization_domains.OrganizationDomainPending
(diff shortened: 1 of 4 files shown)


pkg/organizations/client.go
```diff
@@ -130,21 +136,9 @@ type ListOrganizationsResponse struct {
 	ListMetadata common.ListMetadata `json:"listMetadata"`
 }
 
-type OrganizationDomainDataState string
-
-const (
-	Verified OrganizationDomainDataState = "verified"
-	Pending  OrganizationDomainDataState = "pending"
-)
+type OrganizationDomainDataState = organization_domains.OrganizationDomainDataState
```
mattgd, line 139: I wonder if constant values like this have to be redeclared unfortunately. I don't think they get picked up when defining a type alias.

mattgd (approved):

PR #418, opened 2025-04-07
Add endpoints to interact with WorkOS Vault

hexedpackets:
## Description

Adds a new package for interacting with WorkOS Vault endpoints.

## Documentation

```
[x] Yes
```

https://github.com/workos/workos/pull/37811


pkg/vault/vault_live_example.go
```diff
@@ -0,0 +1,173 @@
...
+func TestCrudObjects(t *testing.T) {
+	setup()
+
+	buf := make([]byte, 16)
+	rand.Read(buf)
+	objectPrefix := hex.EncodeToString(buf)
+
+	objectName := objectPrefix + "-lima"
+	keyContext := KeyContext{"fiber": "Alpalca"}
+	value := "Huacaya 27.7 micron"
```
debussyman, line 25: ❤️

pkg/vault/client.go
```diff
@@ -0,0 +1,550 @@
...
+
+	if c.JSONEncode == nil {
+		c.JSONEncode = json.Marshal
+	}
+}
+
+type KeyContext map[string]interface{}
+
+// Objects
+type ObjectMetadata struct {
```
debussyman, line 66: Is it the right pattern that Vault types are defined in the client, and not extracted into its own file?

hexedpackets: That's how all the other clients are done, but I'll let someone with more experience on the go SDK comment if they feel differently.

debussyman (approved):

kkajla12 (approved):

pkg/vault/client.go
```diff
@@ -0,0 +1,544 @@
...
+	endpoint := fmt.Sprintf("%s/vault/v1/kv", c.Endpoint)
+	req, err := http.NewRequest(http.MethodGet, endpoint, nil)
+	if err != nil {
+		return ListObjectsResponse{}, err
+	}
+
+	req = req.WithContext(ctx)
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "Bearer "+c.APIKey)
+	req.Header.Set("User-Agent", "workos-go/"+workos.Version)
```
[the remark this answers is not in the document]
kkajla12: A good approach would be to have our own `RequestBuilder` that adds common headers/configuration and helps the caller set additional attributes on the request. To not hold your PR up, we can look into this separately.

pkg/vault/client.go
kkajla12: It's an OK pattern for now since none of these types have any instance methods or other logic associated with them. As a type becomes more complex and has instance methods, it would be a good candidate to move into its own file (along with unit tests).

kkajla12 (approved):

PR #426, opened 2025-05-09
Add "warnings" field to FGA query and check response

atainter:
## Description
Adds `warnings` field to the FGA check and query responses. `Warnings` is a slice of `WarningWrapper` that contain a specific warning type. Warning types include:

- `BaseWarning` - catch all for warnings not supported by the current SDK version
- `MissingContextKeysWarning` - missing context keys found while evaluating policies in the check / query request 
- `ConvertSchemaWarning` - validation failed for schema [not used with check / query]

### DIFF
--- a/pkg/fga/client.go
+++ b/pkg/fga/client.go
@@ -380,21 +380,23 @@ type CheckBatchOpts struct {
 }
 
 type Warning interface {
-	Message() string
-	Code() string
+	Warning() string
 }
 
 type BaseWarning struct {
-	C   string `json:"code"`
-	Msg string `json:"message"`
+	Code    string `json:"code"`
+	Message string `json:"message"`
 }
 
-func (b BaseWarning) Code() string    { return b.C }
-func (b BaseWarning) Message() string { return b.Msg }
+func (b BaseWarning) Warning() string { return fmt.Sprintf("%s: %s", b.Code, b.Message) }
 
 type MissingContextKeysWarning struct {
 	BaseWarning
 	Keys []string `json:"keys"`
+}
+
+func (m MissingContextKeysWarning) Warning() string {
+	return fmt.Sprintf("%s: %s [%s]", m.Code, m.Message, strings.Join(m.Keys, ", "))
 }
 
 type ConvertSchemaWarning struct {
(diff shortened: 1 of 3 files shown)


pkg/fga/client.go
```diff
@@ -379,10 +379,79 @@ type CheckBatchOpts struct {
 	WarrantToken string `json:"-"`
 }
 
+type Warning interface {
+	GetWarning() string
```
kkajla12, line 383: It's more idiomatic in Go to name this and the other getter `Warning()` and `Code()`. You can then lowercase the `Message` and `Code` attributes in the implementing struct to "unexport" them so the caller must rely on `Warning()` and `Code()`.

On a separate note, why not rename `Warning()` to `Message()`? That way it doesn't stutter (i.e. `Warning.Warning`).

pkg/fga/client.go
```diff
@@ -379,10 +379,79 @@ type CheckBatchOpts struct {
...
+}
+
 type CheckResponse struct {
-	Result     string    `json:"result"`
-	IsImplicit bool      `json:"is_implicit"`
-	DebugInfo  DebugInfo `json:"debug_info,omitempty"`
+	Result     string           `json:"result"`
+	IsImplicit bool             `json:"is_implicit"`
+	DebugInfo  DebugInfo        `json:"debug_info,omitempty"`
+	Warnings   []WarningWrapper `json:"warnings,omitempty"`
```
kkajla12, line 454: Did you try making `Warnings` something like this?
```go
Warnings []Warning
```
Then implement `MarshalJSON` and `UnmarshalJSON` for each of the structs that implements `Warning`? I believe Go's dynamic dispatch will call the right method.

pkg/fga/client.go
atainter: Oops yes I meant `Message`

pkg/fga/client.go
atainter: I think that works on marshal, but not unmashal because Go does not know the concrete type for each Warning. I think an alternative would be to write a custom unmarshal / marshal for CheckResponse / QueryResponse, but it was duplicating code so I decided against it

stanleyphu (approved):

stanleyphu (approved):

PR #431, opened 2025-06-18
add support for custom oauth scopes

stacurry:
## Description
- adds support for the `oauth_tokens` object in the authenticate API response
- adds support for the new `provider_scopes` query parameter

## Documentation

```
[X] Yes
```

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/google/go-querystring/query"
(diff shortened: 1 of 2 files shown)


pkg/usermanagement/client.go
```diff
@@ -980,6 +1002,10 @@ func (c *Client) GetAuthorizationURL(opts GetAuthorizationURLOpts) (*url.URL, er
 		query.Set("screen_hint", string(opts.ScreenHint))
 	}
 
+	if len(opts.ProviderScopes) > 0 {
+		query.Set("provider_scopes", strings.Join(opts.ProviderScopes, ","))
```
mthadley, line 1006: Should these be space-separated? Similar to [the PHP SDK implementation](https://github.com/workos/workos-php/pull/289/files#diff-e10143672057f6945a897ea4fe63716e89c429029f0bb2a0e3414955e6081a31R695).

```suggestion
		query.Set("provider_scopes", strings.Join(opts.ProviderScopes, " "))
```

stacurry: Ah it is actually the PHP SDK code that is wrong - our API is expecting a comma-separated list. I'll update my PHP PR.

mthadley (approved):

PR #432, opened 2025-06-23
bump to v4.41.0

stacurry:
## Description
version bump that includes:
- https://github.com/workos/workos-go/pull/431

## Documentation

```
[ ] Yes
```

### DIFF
--- a/internal/workos/workos.go
+++ b/internal/workos/workos.go
@@ -2,5 +2,5 @@ package workos
 
 const (
 	// Version represents the SDK version number.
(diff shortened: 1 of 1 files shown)


mthadley:
@stacurry A heads up that I merged https://github.com/workos/workos-go/commit/916bb0c6dcd5edfde3c04c942a87476b456498e9 so that would now also be included in this changeset.

Technically the structure of `ListIdentitiesResult` is different, but it was never correct nor worked, so I don't think I'd consider it a breaking change.

PR #433, opened 2025-06-24
Correctly deserialize `ListIdentitiesResult` as array

blaine-arcjet:
## Description

When trying to use the `usermanagement.ListIdentities` function, the result is always an error. I believe this is because the response is just an array of `Identity` instead of an object containing the `identities` field.

## Documentation

I don't believe this needs docs update because the examples in the List Identities section don't contain a Go example.

### DIFF
--- a/pkg/usermanagement/client.go
+++ b/pkg/usermanagement/client.go
@@ -600,9 +600,7 @@ type RevokeSessionOpts struct {
 	SessionID string `json:"session_id"`
 }
 
-type ListIdentitiesResult struct {
-	Identities []Identity `json:"identities"`
-}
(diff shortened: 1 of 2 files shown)


mthadley (approved):
Thanks for noticing and fixing!

I pushed up a quick PR to fix some now outdated tests but otherwise looks great.

blaine-arcjet:
Excellent, thanks! Hopefully a release can roll out so I can use this in some work I'm doing 🙏

gcarvelli:
@blaine-arcjet Your change is available in [v4.41.0](https://github.com/workos/workos-go/releases/tag/v4.41.0) of the sdk. Thanks for your contribution!

PR #439, opened 2025-07-10
Update metadata serialization for updating organizations

awolfden:
## Description
Update metadata serialization for updating organizations. Coreweave flagged that this was broken and not in line with the update user metadata serialization. 

## Documentation

```
[ ] Yes
```

### DIFF
--- a/pkg/organizations/client.go
+++ b/pkg/organizations/client.go
@@ -426,9 +426,12 @@ func (c *Client) UpdateOrganization(ctx context.Context, opts UpdateOrganization
 		//
 		// Deprecated:  Use DomainData instead.
 		Domains []string `json:"domains,omitempty"`
+
+		// The Organization's metadata.
+		Metadata map[string]string `json:"metadata,omitempty"`
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/organizations/client.go
```diff
@@ -426,11 +426,14 @@ func (c *Client) UpdateOrganization(ctx context.Context, opts UpdateOrganization
 		//
 		// Deprecated:  Use DomainData instead.
 		Domains []string `json:"domains,omitempty"`
+
+		// The Organization's metadata.
+		Metadata map[string]string `json:"metadata,omitempty"`
```
mattgd, line 431: I noticed we have two `UpdateOrganizationOpts` definitions, and one of them already had metadata. Are we able to remove this struct and use the existing one? Then I'm pretty sure we can just pass opts into `c.JSONEncode` directly.

awolfden: Good call! Looks like this is how we do it for updateuser too so now with this update it will follow the same pattern.

mattgd (approved):

PR #444, opened 2025-07-30
Expose delete organization_domain endpoint - DAAP-1754

ajworkos:
## Description

Exposes an SDK method to delete an organization domain

## Documentation

```
[x] Yes
```

Docs PR:
https://github.com/workos/workos/pull/42288

### DIFF
--- a/pkg/organization_domains/client_test.go
+++ b/pkg/organization_domains/client_test.go
@@ -260,3 +260,65 @@ func verifyDomainTestHandler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 	w.Write(body)
 }
+
+func TestDeleteDomain(t *testing.T) {
+	tests := []struct {
+		scenario string
+		client   *Client
+		options  DeleteOrganizationDomainOpts
+		err      bool
+	}{
+		{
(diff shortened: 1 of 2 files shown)


pkg/organization_domains/client.go
```diff
@@ -204,3 +209,39 @@ func (c *Client) VerifyDomain(ctx context.Context, opts VerifyOrganizationDomain
 	err = dec.Decode(&body)
 	return body, err
 }
+
+// DeleteDomain deletes an Organization Domain.
+func (c *Client) DeleteDomain(ctx context.Context, opts DeleteOrganizationDomainOpts) error {
```
mattgd, line 214: I see this is still in draft, but could we add a test for `DeleteDomain`?

ajworkos: Thanks! Added

mattgd: Also realized, should this be called `DeleteOrganizationDomain`?

ajworkos: Looks like `DeleteDomain` aligns with the existing names in this module of `VerifyDomain`, `CreateDomain`, and `VerifyDomain`

PR #485, opened 2026-01-16
retryablehttp: add test for context cancellation during retry backoff

Falasefemi2:
This PR adds a test to verify that HttpClient.Do stops retrying
and returns context.Canceled if the request context is canceled
during the backoff sleep between retries. 

It ensures the correct behavior and prevents potential goroutine/timer leaks.


gjtorikian:
Thanks for your contribution! The ideas here (context cancellation during backoff sleep) are now present in v7 of the SDK.

PR #524, opened 2026-04-07
feat!: generate Go SDK from OpenAPI spec

gjtorikian:
## Description

This branch introduces the v7 Go SDK surface behind a single root `workos` package and shared `workos.Client`.

It also:
- regenerates the API surface and models from the spec
- removes the legacy `pkg/*` service packages in favor of `client.Service()` accessors
- updates list, error, and webhook patterns to match the new client shape
- adds `docs/V7_MIGRATION_GUIDE.md` to document consumer migration

## Documentation

```
[ ] Yes
```


swaroopAkkineniWorkos:
@greptile review

passwordless.go
```diff
@@ -0,0 +1,58 @@
...
+
+package workos
+
+import (
+	"context"
+	"fmt"
+)
+
+// passwordlessService handles Passwordless session operations.
+type passwordlessService struct {
```
swaroopAkkineniWorkos: passwordless.go:16 and vault_kv.go:16  both initialize the client in the separate file, vs in workos.go like the other clients

user_management.go
```diff
@@ -0,0 +1,1072 @@
...
+	_, err := s.client.request(ctx, "GET", fmt.Sprintf("/sso/jwks/%s", clientID), nil, nil, &result, opts)
+	if err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
+
+// UserManagementCreateAuthenticateParams contains the parameters for CreateAuthenticate.
+type UserManagementCreateAuthenticateParams struct {
+	Body interface{} `json:"-"`
```
swaroopAkkineniWorkos: is there a reason for the untyped blog here for the entire body?

authorization.go
```diff
@@ -0,0 +1,581 @@
...
+	PaginationParams
+	// PermissionSlug is the permission slug to filter by. Only child resources where the organization membership has this permission are returned.
+	PermissionSlug string `url:"permission_slug" json:"-"`
+	// ParentResourceID is the WorkOS ID of the parent resource. Provide this or both `parent_resource_external_id` and `parent_resource_type_slug`, but not both.
+	ParentResourceID *string `url:"parent_resource_id,omitempty" json:"-"`
+	// ParentResourceTypeSlug is the slug of the parent resource type. Must be provided together with `parent_resource_external_id`.
+	ParentResourceTypeSlug *string `url:"parent_resource_type_slug,omitempty" json:"-"`
+	// ParentResourceExternalID is the application-specific external identifier of the parent resource. Must be provided together with `parent_resource_type_slug`.
+	ParentResourceExternalID *string `url:"parent_resource_external_id,omitempty" json:"-"`
+}
```
swaroopAkkineniWorkos: it would be cool if we could put types like (ParentResourceTypeSlug + ParentResourceExternalID) into it's own object so they are grouped together by default. This comes up a lot across the sdk

authorization.go
```diff
@@ -0,0 +1,827 @@
...
+
+// AuthorizationListResourcePermissionsParams contains the parameters for ListResourcePermissions.
+type AuthorizationListResourcePermissionsParams struct {
+	PaginationParams
+}
+
+// ListResourcePermissions list effective permissions for an organization membership on a resource
+// Returns all permissions the organization membership effectively has on a resource, including permissions inherited through roles assigned to ancestor resources.
+func (s *AuthorizationService) ListResourcePermissions(ctx context.Context, organizationMembershipID string, resourceID string, params *AuthorizationListResourcePermissionsParams, opts ...RequestOption) *Iterator[AuthorizationPermission] {
```
swaroopThereItIs-workos, line 188: It seems odd to have the path params be passed in separately than the body params. I think the dev shouldn't really care about that. I think the dev should have to pass in one request body and we handle the rest. Curious what you think?

swaroopAkkineniWorkos (approved):

authorization.go
gjtorikian: So, I "trained" the SDK autogenerator's on Stripe's SDKs, and this is the pattern they follow. I kind of prefer keeping them as positional args because:

1. They're required and non-nullable (enforced at compile time)
2. It shows "where am I operating?" vs "what am I sending" / "how do I filter?"

